PR #9, opened 2020-09-02
re-theme: hookFactory update

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-280)

## Context

* There's a bug in hookFactory where it does not update the `activeValue`. which causes the button component to not update its styles on **hover** and **active** state
* keg-comp

## Goal

* Update `hookFactory` so that it updates `activeValue` whenever there is a change to it
* also update keg-components Button, so that it merges the styles for active and hover states

## Updates

- updated button component to **merge** the styles for hoverStyles and activeStyles
- updated re-theme: **hookFactory** to update the activeValue when it changes

## Testing

* run `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-280-groupbooking-modal-footer web`
* navigate to http://kegdev.xyz/test
* open one of the `Group Booking Modal` 1,2 or 3
* resize your browser to mobile size < 480px
* see that the fontSize of the button changed
* hover over either buttons, verify that the fontSize remains the same as the default

### Unit test
* Pull the branch `gh pr checkout 9`
* run `yarn test` on keg-components and re-theme to verify that it still passes

### DIFF
--- a/repos/keg-components/src/components/button/button.wrapper.js
+++ b/repos/keg-components/src/components/button/button.wrapper.js
@@ -53,13 +53,13 @@ export const ButtonWrapper = props => {
   const [ hoverRef, hoverStyles ] = useThemeHover(
     get(btnStyles, 'default', {}),
     get(btnStyles, 'hover'),
-    { ref, noMerge: true }
+    { ref, noMerge: false }
   )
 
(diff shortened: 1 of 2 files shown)


lancetipton (approved):
Looks good. All tests pass. Thanks

PR #10, opened 2020-09-03
updating @svkeg to @keg-hub 

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-359)

## Context

- our official npm org is @keg-hub . we should slowly start publishing our packages there instead of where it is now (@svkeg)
- since our tap-events-force is using keg-core as npm package. i figured i would start there

## Goal

* main goal was to publish `keg-core` to @keg-hub npm
* ended up publishing most of the other repos that `keg-core` depends on
* packages link: https://www.npmjs.com/org/keg-hub
* packages updated/published:
    - keg-core
    - keg-components
    - re-theme
    - jsutils
    - tap-resolver
    - spawn-cmd
    - args-parse
    - ask-it
    

## Updates
* `/repos/*`
   * replaced `@svkeg` with `@keg-hub`

* `/taps/tap-release-client`
   * replaced `@svkeg` with `@keg-hub`

## Testing

* Pull this branch
* do a clean install on `keg-cli`
     * `cd repos/keg-cli`
     * `yarn clean:full`
* then you can theoretically do an install on all the repos at once
    * `keg hub run install`
* run test for all the repos
    * `keg hub run test` or i suggest running `keg hub run test <repo_name>` so it's easier to track which one passes/fails if any

### Taps
* navigate back to the hub `keg hub`
* navigate to the tap-release-client. `cd repos/tap-release-client`
* run `yarn install`
* run `yarn web`
* verify that you see the application running


lancetipton:
One other thing that might help. 
If you run `keg cli print <name of thing>`, it will print out the values the cli has for what ever `name of thing` is
For example `keg cli print doc`
  * will print out all the `docker constants`; so you can see what's available to you
Or `keg cli print core`
  * will print out the constants it uses to build the `keg-core`
Or `keg cli print --context task --task docker.exec`
  * will print the task definition for a task. In this case the `docker.exec` task

mpcarolin (approved):
Looks good and worked for me.

I could not run the global repo commands like `keg hub run install`, but that's unrelated to your changes here. I think we should add a `--synchronous` flag that runs each repo's command one at a time, rather than all in parallel, because I think the parallel execution is causing problems.

Anyway, installing and testing each repo one at a time, everything worked and passed the unit tests. I also ran `yarn web` for core and that behaved as expected.

lancetipton (changes requested):
This all worked for me, but you missed the `tap-release-client` in the `/taps` folder. Can update that one as well. 

Thanks

lancetipton (approved):
Nice, thanks dude. Every thing looks to be working as expected.

PR #11, opened 2020-09-03
ZEN-349: Expose Localhost / Tunneling

mpcarolin:
**Ticket**: [ZEN-349](https://jira.simpleviewtools.com/browse/ZEN-349

## Context

* We cannot currently test `kegdev.xyz` on native mobile devices, or any machine other than the dev machine on which the keg-cli resides
* We can use public tunnels to fix this

## Goal

* implement a cli task for creating public tunnels to `kegdev.xyz`, accessible on a mobile device

## Updates

* `repos/keg-cli/package.json`
  * minor version bump of the version
  * added `ngrok` and `qr-terminal` dependencies
* `repos/keg-cli/src/tasks/docker/docker.js`
  * my task is in the `docker` scope, so I required it here
* `repos/keg-cli/src/tasks/docker/tunnel.js`
  * the task for tunneling

## Testing a docker package
* `keg hub`
* `keg pr 11`
* `keg cli`
* `yarn install`
* `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-317-rollup-export-truth`
  * this starts up the tap
* verify you can visit `kegdev.xyz`, then close that tab
* open a new tab in your terminal, then run `keg d tunnel`
* this will create an ngrok tunnel to the docker ip, and it will print out a tunnel url and a qr code
* try visiting the tunnel link printed out at the top. Ensure it hosts the tap.
* use an iphone or android device's camera to view the qr code printed out. Follow the link that appears as a notification, then verify that you can view the tap on your mobile device.
* kill the terminal process for the tunnel, and verify that you can no longer visit that tunnel url

## Test local
* start up anything on localhost (not in docker), such as keg-components
* `keg d tl port=<port of your local app>`
* navigate to the tunnel url it prints out
* verify you can view your app at that url

### DIFF
--- a/repos/keg-cli/src/tasks/docker/tunnel.js
+++ b/repos/keg-cli/src/tasks/docker/tunnel.js
@@ -1,6 +1,7 @@
 const ngrok = require('ngrok')
 const qrcode = require('qrcode-terminal')
 const { Logger } = require('KegLog')
+const { DOCKER } = require('KegConst/docker')
 
 /**
  * Create a public tunnel the the url (by default, to `kegdev.xyz`)
@@ -15,9 +16,9 @@ const tunnel = async args => {
   const { params } = args
   const { qr, source, port } = params
 
-  const dockerIP = process.env.KEG_DOCKER_IP
+  const dockerIP = DOCKER.PREFIXED.KEG_DOCKER_IP
   const srcUri = source || port || dockerIP
-
+  
   Logger.header('Starting tunnel...', 'white')
 
   const tunnelUrl = await ngrok.connect(srcUri)
@@ -32,14 +33,15 @@ const tunnel = async args => {
 module.exports = {
   tunnel: {
     name: 'tunnel',
-    alias: ['t', 'tl', 'ngrok' ],
+    alias: [ 't', 'tl', 'tn', 'tnl', 'tun', 'ngrok' ],
     action: tunnel,
     description: 'Creates a public tunnel to the running docker container available at kegdev.xyz',
     example: 'keg docker tunnel <options>',
     options: {
       qr:  {
         description: 'if true, `tunnel` will display a qr code that encodes the tunnel\'s public uri. This can make connecting a mobile device to the tunnel much easier.',
-        example: 'keg docker tunnel --qr',
+        example: 'keg docker tunnel --qr false',
+        default: true,
       },
       source: {
         description: "optional override for the source url or ip of the tunnel. By default, this command tunnels kegdev.xyz",
--- a/repos/keg-cli/src/tasks/docker/tunnel.js
+++ b/repos/keg-cli/src/tasks/docker/tunnel.js
@@ -1,0 +1,54 @@
+const ngrok = require('ngrok')
+const qrcode = require('qrcode-terminal')
+const { Logger } = require('KegLog')
+
+/**
+ * Create a public tunnel the the url (by default, to `kegdev.xyz`)
+ * @function
+ * @param {Object} args - arguments passed from the runTask method
+ * @param {string} args.command - Initial command being run
+ * @param {Array} args.params - arguments passed from the command line
+ *
+ * @returns {void}
+ */
+const tunnel = async args => {
+  const { params } = args
+  const { qr, source, port } = params
+
+  const ip = process.env.KEG_DOCKER_IP
+
+  Logger.header('Starting tunnel...', 'white')
+
+  const url = await ngrok.connect(source || port || ip)
+
+  Logger.spacedMsg('Source:', Logger.color('magenta', ip))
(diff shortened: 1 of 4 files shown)


repos/keg-cli/src/tasks/docker/tunnel.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param {string} args.command - Initial command being run
+ * @param {Array} args.params - arguments passed from the command line
+ *
+ * @returns {void}
+ */
+const tunnel = async args => {
+  const { params } = args
+  const { qr, source, port } = params
+
+  const ip = process.env.KEG_DOCKER_IP
```
lancetipton: While this is not wrong, there's a better way to do this. 
```js
// import the docker constants at the top of the file
const { DOCKER } = require('KegConst/docker')

// Then on this line set it like this
const ip = DOCKER.MACHINE.IP
// Or like this, because at some point we'll switch to docker for desktop
const ip = DOCKER.PREFIXED.KEG_DOCKER_IP
```
Most likely when we switch to docker for desktop, this will have to be updated anyways, so either option is fine

repos/keg-cli/src/tasks/docker/tunnel.js
```diff
@@ -0,0 +1,54 @@
...
+
+  Logger.header('Starting tunnel...', 'white')
+
+  const url = await ngrok.connect(source || port || ip)
+
+  Logger.spacedMsg('Source:', Logger.color('magenta', ip))
+  Logger.spacedMsg('Tunnel:', Logger.color('green', url))
+
+  // show a qr code on the terminal that the user can use to quickly access the tunnel on a mobile device
+  qr && qrcode.generate(url, { small: false })
```
lancetipton: The qr code is dope. So much so, that I would make it turned on by default. Then you would not need to pass `--qr` everytime

In the `options object` of the `task definition` down below; you can set a `default` for the option.
See my comment in that section to see what I'm saying

repos/keg-cli/src/tasks/docker/tunnel.js
```diff
@@ -0,0 +1,54 @@
...
+
+module.exports = {
+  tunnel: {
+    name: 'tunnel',
+    alias: ['t', 'tl', 'ngrok' ],
+    action: tunnel,
+    description: 'Creates a public tunnel to the running docker container available at kegdev.xyz',
+    example: 'keg docker tunnel <options>',
+    options: {
+      qr:  {
```
lancetipton, line 40: You can add a `default` key / value here for options and I would set `qr` to `true` by default
Looks like this => 
```js
      task: {
        description: `If true, "tunnel" will display a qr code that encodes the tunnel\'s public uri. This can make connecting a mobile device to the tunnel much easier.`,
        example: 'keg docker tunnel --qr false',
        default: true
      }
```

lancetipton (changes requested):
Looks good, didn't test it cause I'm in the middle of something; but I will.
There's a few comments I left. Let me know if you have questions. Thanks again

lancetipton:
One other thing that might help. 
If you run `keg cli print <name of thing>`, it will print out the values the cli has for what ever `name of thing` is
For example:
* `keg cli print doc`
   * will print out all the `docker constants`; so you can see what's available to you
* `keg cli print core`
   * will print out the constants it uses to build the `keg-core`
* `keg cli print task docker.exec`
   * will print the task definition for a task. In this case the `docker.exec` task

daniel112 (approved):
nice. works really well for me

repos/keg-cli/src/tasks/docker/tunnel.js
```diff
@@ -0,0 +1,54 @@
...
+  Logger.spacedMsg('Tunnel:', Logger.color('green', url))
+
+  // show a qr code on the terminal that the user can use to quickly access the tunnel on a mobile device
+  qr && qrcode.generate(url, { small: false })
+}
+
+module.exports = {
+  tunnel: {
+    name: 'tunnel',
+    alias: ['t', 'tl', 'ngrok' ],
```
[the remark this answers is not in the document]
mpcarolin: sounds good, updated with a few more too

repos/keg-cli/src/tasks/docker/tunnel.js
mpcarolin: I was wondering if something like that existed. Thanks, just updated

repos/keg-cli/src/tasks/docker/tunnel.js
mpcarolin: Yeah works for me. Updated

repos/keg-cli/src/tasks/docker/tunnel.js
mpcarolin: 👍  updated

PR #14, opened 2020-09-07
Zen 353 components data class

lancetipton:
**Ticket [ZEN-354](https://jira.simpleviewtools.com/browse/ZEN-354)** 

**Notes**
* Originally this ticket was to add a `dataSet.class` attribute to the components
  * This has been changed to instead use the `className` prop
  * This in more consistent with other web platforms
  * On Native, it is ignored

## Context
* Events-Force is defining their styles in CSS
* Their clients can define custom CSS as well
* Currently, the tap, using re-theme, is only compatible with JSON style objects
* We need to be able to pull styles from stylesheets already applied to the DOM and apply them to keg-components

## Goal
* Update all keg-components to accept a className prop
  * Using the `useClassname` and `useClasslist` hooks, the `className` can be used instead of the `dataSet`
* Add a default `className` prop to all primitive components
* Merge the passed in `className` prop with the default `className` prop

## Updates
**Global**
  * Removed `@expo/vector-icons` && `react-native-vector-icons`
  * Added `react-native-svg` && `react-native-svg-web`
  * Added `svgr cli` to convert SVG's to React Native Components
  * Added scripts helpers to do SVG conversion
* `repos/keg-components/README.md`
  * Updated with more information
  * Added notes about updated / added code
* `repos/keg-components/configs/aliases.json`
  * Cleaned up to match code updates
* `repos/keg-components/configs/jest.config.js`
  * Removed `@expo/vector-icons`
* `repos/keg-components/setupTests.js`
  * Added mocks for `react-native-svg` to `react-native-svg-web`
**Assets/Icons**
* `repos/keg-components/src/assets/icons/**`
  * Added a number of svg icons as react components
    * Mostly used in stories, but also used in some places of the code
* `repos/keg-components/src/assets/icons/svgIcon/svgIcon.js`
  * Added SVG component for consistent SVG icon rendering
* `repos/keg-components/src/assets/spinners/**`
  * Removed spinners as they are no longer being used

**Components**
* `repos/keg-components/src/components/**`
  * Updated to use `useClassList` and `useClassName` hooks as needed
  * Updated to use `SVG` component as needed
  * Added a **Touchable** component
    * Replaced components that were using `TouchableOpacity` and `TouchableNativeFeedback`
  * Added **accessibilityRole** to components where needed
    * This helps with symantec HTML, as well Dom element type to be used 
  * Removed Component wrapper in cases where no longer needed
    * `Button` / `Form` components
  * Added `createElement.js` component
    * Allows creating custom Components using the `react-native-web` framework
  * Fixed `Checkbox` component styles
    * Added styles are defined here so that they are enforced
    * Due to how the checkbox is designed, these styles should not be changed 
    * Which is why they are not defined within the theme
  * Added helpers to `noop`
    * `noPropObj` && `noPropArr`
    * Replace locations in components where they were just being set to an empty object or array
  * Fixed bug with TouchableIcon causing the wrapper styles to not be used
  * Updated `style` prop in a number of locations to use the `array` syntax
**Theme**
  * Updated theme defaults to allow being overwritten
  * Updated all references to `defaults.json` to use `getThemeDefaults` instead 
  * Added a `setThemeDefaults` that accepts a `defaults` object
    * **IMPORTANT** - This should be called prior to theme initialization
      * This way the defaults will be updated before the theme is built
**Keg-Core**
* `app.json`
  * Added mapping of `react-native-svg` to `react-native-svg-web` on web platform

## Testing
* **TODO**


lancetipton:
Closing, these changes are now part of #20

PR #15, opened 2020-09-07
Migrate to docker-desktop

lancetipton:
## IMPORTANT
* **THESE CHANGES WILL NOT EFFECT ANYONE WHO DOES NOT MIGRATE**
* These changes to the `keg-cli` allow it to work with either `docker-desktop` or `docker-machine`
* **But**, only do the tests if you want to **MIGRATE** to `docker-desktop`
  * Otherwise you will have to re-build your `docker-machine` box from scratch
    * Not a big deal, but is a pain if you don't want to use `docker-desktop`
    * Seriously though, `docker-desktop` is **WAY** smoother, and faster
* After the migration, even switching to a different branch will not cause issues
  * The key is the `KEG_NO_MACHINE` env. Removing that will revert back to `docker-machine`
    * This will cause all kinds of error, because `docker-machine` has been removed after this migration
    * So don't do that!

## Context
* Ya'll have asked for using **Docker for Desktop** with the Keg-CLI
* **Docker-Machine** works but there are a number of downsides
  * Requires Virtual Box
  * Sometimes freezes for no apparent reason
  * Docker has offically stopped development of it
* So we would need to migrate to **Docker for Desktop** eventually

## Goal
* Migrate the Keg-CLI to **Docker for Desktop**
* There should be no loss in functionality
* Provide a way for mapping ports of running docker container
  * Mac - `dnsmasq`
  * Windows - No option currently but it is possible
  * Will always work on `localhost`; i.e. `127.0.0.1`

## Updates
* Added some Caddy ENVs, these are for `Keg-Proxy` which is still a work in progress
  * They can be ignored for now
* `repos/keg-cli/keg`
  * Update to check if using `docker-machine` or `docker-desktop`
* `repos/keg-cli/scripts/docker/dnsmasq.conf`
  * This is the basic config for `dnsmasq.conf`
  * Allows docker container ip and port mapping
  * This allows for functionality similar to `docker-machine`
* `repos/keg-cli/scripts/docker/dnsmasq.sh`
  * Mac only `dnsmasq` install / setup script
  * Sets up an local `dns` server to allow domain ip resolution locally
  * Follow the instructions in the **Testing** section below to use
* `repos/keg-cli/scripts/docker/docker-desktop.env`
  * Docker envs when using `docker-desktop`
* `repos/keg-cli/scripts/docker/mac-desktop.sh`
  * Mac only `docker-desktop` setup script
  * Run this script to migrate from `docker-machine` to `docker-desktop`
  * Follow the instructions in the **Testing** section below to use
* `repos/keg-cli/src/constants/docker/machine.js`
  * Update to use correct docker envs based on `KEG_NO_MACHINE` env
* `repos/keg-cli/src/utils/builders/buildContextEnvs.js`
  * By default `docker-compose` creates `services` on their own network
  * This is a non-issue in `docker-machine` because they are all running on the same VM
    * This allows them to communicate through the VMs networking bridge
  * There is no access to the VM in `docker-desktop`, so we must work around this
    * Added `COMPOSE_PROJECT_NAME` env to the contextEnvs
      * This ensure all `services` started by `docker-compose` are on the same network
      * This allows `dnsmasq`, and other docker containers to communicate with eachother

### Unrelated Updates
**Notes**
  * Most of these changes come for work being done on `keg-proxy`. It's still a work in progress, so it has not been added to the `keg-hub` yet, but these changes are still beneficial and should be added to the `keg-cli`

* `repos/keg-cli/src/tasks/docker/image/run.js`
  * Added `entrypoint` option to allow overwriting the default entry point on image run
* `repos/keg-cli/src/tasks/docker/prune.js`
  * Updated `docker prune` task
    * **IMPORTANT** - This affects the command `keg d clean`
  * The task no longer fully removes everything in docker when run
    * You now must pass the `all` option to get the same functionality
    * Example `keg d clean` === `keg d clean --all`
  * Added force option
    * Passing force true will add all `docker prune` options
      * `-a`, `-f`, `--volumes`
* `repos/keg-cli/src/tasks/tap/run.js`
  * Removed `entry` as an alias for `cmd` option
    * `entry` is now it's own option to override the `ENTRYPOINT` of a dockerfile
* `repos/keg-cli/src/utils/builders/buildBaseImg.js`
  * Added ability for `taps` to not require the `Keg-Base Docker Image` be built
    * Taps must explicitly set `KEG_FROM_BASE` to `false` in their values file
      * Doing this will bypass auto-creating the `Keg-Base Docker Image`
    * Can be used in both the `Re-Theme` tap, and the `Visitapps` tap
    * I also use it for a number of my own taps
  * Added ability for `taps` to not automatically create a sync between localhost and the container
  * Taps must explicitly set `KEG_AUTO_SYNC` to `false` in their values file
  * Can be used for `Taps` and injected apps that don't require a mutagen sync when run

## Testing

**Setup**
* Pull this branch
  * `keg hub && keg pr 15`
* Remove all the things `docker-*`
  * Remove containers
    * Run `keg dc` - See list of all containers
    * Run `keg dc rm`
      * Then select container to remove
      * Keep removing containers until there are none left
  * Remove images
    * Run `keg d clean -f`
    * Run `keg di` and ensure no images are left
* Anything not removed will just take up space on your hard-drive
  * You will not be able to access it from `docker-desktop`
  * **So be sure you clean everything before migrating**

**Test 1** - Migrate to `docker-desktop`
* Run the bash script => `repos/keg-cli/scripts/docker/mac-desktop.sh`
  * `bash ~/keg-hub/repos/keg-cli/scripts/docker/mac-desktop.sh`
    * This script should do the following
      * **Part 1** - Remove `docker`, `docker-compose`, `docker-machine`
        * Before removing `docker-machine` it should remove the `docker-keg` VM
        * After the script runs
          * Check to ensure `docker-keg` was removed from VirtualBox
          * VirtualBox can be opened from your `/Applications` folder
          * If it still exists remove it, **but it should not exist**
      * **Part 2** - Next it will install `docker-desktop`
        * First it will download the `docker-desktop.dmg`
        * Mount it as a `volume`
        * Then install the `Docker.app` into your `/Applications` directory
        * Then un-mount the `volume`
          * Sometimes the `volume` does not get unmounted
          * This is a bug with the Mac OS, not much I can do from code
          * So just open Finder and check the `Locations` Section on the right
            * If it didn't un-mount, you will see it next to `iCloud Drive`
            * Just click the `^` button to un-mount it `( Not a big deal if it fails )`
      * **Part 3** - It should add the env `KEG_NO_MACHINE` to your `.bashrc` or `.bash_profile`
        * After the script runs
          * Check to ensure the `KEG_NO_MACHINE` env was added to either one of them
          * Either open the files, or run `cat ~/<file name> | grep KEG_NO_MACHINE` in the terminal
            * You should see it print out `export KEG_NO_MACHINE=true` for one of the files
        * The correct file depends on your host machine setup
          * By default it tries to add it to your `~/.bash_profile`
          * If that file does not exist, then it tries to add it to `~/.bashrc`
          * If it does not exist, you can add it manually by adding `export KEG_NO_MACHINE=true` to your `~/.bash_profile`
    * After the script finishes it should print
      * `Finished with exit code 0`
        * If it prints anything other then exit code `0`, then there was an issue
        * Look back through the logs to try and find the error
        * Then try to investigate the issue.
          * Report it back to me, so I can update as needed
      * It should also print out a pretty message, letting you know the migration worked
* If everything went well, then you are now using `docker-desktop`
  * **Open a new terminal window**, then run `keg dc`
    * It should print out like normal **with out** error
    * Run the command `docker-machine` it should throw an error
      * This is expected, as `docker-machine` should no longer exist
    * Run `docker info | grep "Operating System"`
      * It should print out `Operating System: Docker Desktop`

**Test 2** - Add share folders
* Open the `docker-desktop menu`, then select `preferences`
  * The docker-desktop try icon should be in the top right screen, next to the wi-fi icon
    * Here's a pic => http://snpy.in/Vpxk7Z
  * Click on the `Resources` tab, then `FILE SHARING`
    * Then add the `~/keg-hub` and `~/.kegConfig` directories to allow docker to access these folders
    * Here's a pic => http://snpy.in/81VbT1
* Next start any keg app or tap with the keg-cli
  * For example => `keg components start`
    * This should build and run the `keg components` Storybook within a docker container just like before
    * Ensure docker should **NOT** show any folder permissions / sharing errors
  * Once started Open your browser at `http://localhost`
* **IMPORTANT** - `docker-desktop` network does not work the same as `docker-machine`
  * Which means `http://kegdev.xyz` does not work
  * Due to this, you must use `http://localhost` or `http://local.kegdev.xyz`  instead of `http://kegdev.xyz`
  * Ensure the app runs as expected
    * File syncing, keg-cli `sync actions`, `keg components att`, should all still work the same

**Test 3** - Install / Setup dnsmasq (optional)

> **UPDATED**
> **This no longer needed!**
> The DNS has been updated to use [http://local.kegdev.xyz](http://local.kegdev.xyz)
> Because of this, setting up `dnsmasq` is only helpful if you want to work **offline** and still resolve `http://kegdev.xyz`
> Moving forward use [http://local.kegdev.xyz](http://local.kegdev.xyz)  **NOT** `http://kegdev.xyz`

* **Setup**
  * This is **ONLY** needed when running `docker-desktop` **AND**
    * You don't want to use `localhost`
    * You don't want to use [http://local.kegdev.xyz](http://local.kegdev.xyz) || [http://*.local.kegdev.xyz](tap.local.kegdev.xyz)
      *  [http://local.kegdev.xyz](http://local.kegdev.xyz) has been setup to resolve to `127.0.0.1`, which allows the access to the docker containers

* **Step 1**
  * Run this script in the terminal
    * `bash <keg-cli path>/scripts/docker/dnsmasq.sh` 
  * Should install dnsmasq
  * Should setup a LaunchDaemon to ensure dnsmasq runs on boot
  * Should create a file as `/Users/<user name>/.kegConfig/dnsmasq.conf`
  * Should add the custom config to the root config at `/usr/local/etc/dnsmasq.conf`
* **Step 2**
  * The DNS resolvers should be updated automatically
  * But you can validate it by check your DNS settings in system preferences
  * Go to system preferences > network
  * Click on the advanced button
  * Then click on the DNS tab
    * Ensure `127.0.0.1` is the first entry in the DNS servers list
    * Looks like this => http://snpy.in/vHkGoj


daniel112 (approved):
I really dig the super detailed Test steps.
everything works as expected.

also tested sync start on:
- `keg-components` sync retheme
- `evf` sync core, retheme, components

PR #19, opened 2020-09-09
Syntax error fix

mpcarolin:
## Context

* When using `keg-components` in a development workflow inside docker, with the `yarn dev` command, the optional chaining operator `?.` was not getting compiled

## Goal

* fix this issue
* also add a PR template at the root of the monorepo, since there isn't one

## Updates

* `.github/PULL_REQUEST_TEMPLATE.md`
  * added the template
* `repos/keg-components/configs/babel.config.js`
  * just updated the name of the babel preset-env preset to be `preset-env` instead of `env`. This does nothing different, since babel prepends `preset` to any preset without that prefix already, but it's more clear
* `repos/keg-components/package.json`
  * added `NODE_ENV=production` to the dev command, so that `?.` gets compiled

## Testing

* `keg pr 19`
* `keg components`
* `yarn dev`
* open `build/esm/kegComponents.native.js` or any other build file
* search the file for any instance of `?.`. Verify there are zero such instances.

### Container testing
* `keg core start`
* `keg core sync components:install:start`
* `keg core start`
* verify that you can open the keg test page at `kegdev.xyz` and you see no errors about the syntax in the terminal output

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,32 @@
+**Ticket**: [issue_id](link_to_jira_ticket)
+
+## Context
+
+* Why was this ticket created?
+
+## Goal
+
+* What do you intend to achieve by completing this task?
+
+## Updates
+
+* List the updates you made to the codebase. Use your best judgement on how detailed the description should be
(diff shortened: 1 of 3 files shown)


lancetipton (approved):
Looks good, thanks

PR #24, opened 2020-09-11
update url to local

daniel112:
## Context

* after keg-hub docker desktop update
* after running any container. it still prints this text out
     * image
* we cant hit that url
* it should be `<context>.local.kegdev.xyz`

## Goal

* update the constant so it displays the correct url

## Updates

* `repos/keg-cli/src/constants/constants.js`
    * updated the url to `http://tap.local.kegdev.xyz/`

## Testing

* pull this branch
* Run a container i.e:
    * `keg evf start`
    * `keg components start`
* see that the served url is something like 
   * `http://<context>.local.kegdev.xyz/`
   * image
* verify that you can navigate to that url

### DIFF
--- a/repos/keg-cli/src/constants/constants.js
+++ b/repos/keg-cli/src/constants/constants.js
@@ -100,7 +100,7 @@ module.exports = deepFreeze({
   ],
 
   // URLs for navigating to the application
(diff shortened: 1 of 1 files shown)


lancetipton (approved):
ohh, nice. Thanks dude

PR #32, opened 2020-09-15
ZEN-278: Hub Updates

mpcarolin:
**Ticket**: [ZEN-278](https://jira.simpleviewtools.com/browse/ZEN-278)

## Context/Goal

* For the group booking modal of ZEN-278, I needed to make a number of updates to keg-hub
* This PR is paired with [this PR for tap-events-force](https://github.com/simpleviewinc/tap-events-force/pull/63)

### `jsutils`

`[semver-status]: minor`

* `repos/jsutils/src/array/omitElement.js`
  * added a simple helper shortcut for returning an array without a specific element. Can work with reference types too, provided that you pass a function as the third argument to select a primitive property for comparing

* `repos/jsutils/src/array/uniqArr.js`
  * updated `uniqArr` to also work with reference types, so long as you pass in a selector function for comparison.

* `repos/jsutils/src/method/noOp.js`
  * added a noOp function

### `keg-cli`
`[semver-status]: minor`
* updated the tap sync configuration to allow syncing jsutils into a tap container

### `keg-components`
`[semver-status]: minor`

* *Summary*: updated checkbox with new disabled features, added a checkbox group component

* `repos/keg-components/src/components/form/checkGroup/checkGroup.js`
  * implemented the checkgroup component, which is a essentially a list of checkboxes with a header. It also includes an optional header checkbox for select-all.

* `repos/keg-components/src/components/form/checkbox/checkbox.js`
  * added disabled styling to the checkbox, including some opacity dimming and a `not-allowed` cursor on web
  * added `disableCheck` and `disableUncheck` props that disable the respective behaviors
  * updated the storybook stories with more examples
  * added support for passing a callback that can provide the consumer the checkbox's setChecked function
  * updated to accept a `CheckIcon` prop for overriding the default checkmark icon
  * updated to allow pressing the side-components to toggle the checkbox (`allowAdjacentPress`)

### `keg-core`
`[semver-status]: minor`
* *Summary*: added new hooks

* `repos/keg-core/core/base/hooks/state/useExternalSet.js`
  * hook for getting a set-like interface to an externally-defined react-state array
  * it's a helper for `useSet`
* `repos/keg-core/core/base/hooks/state/useSet.js`
  * stores elements in react state, while also ensuring no duplicates
  * returns an object with the same interface as a JS Set
* `repos/keg-core/core/base/hooks/store/useStoreItems.js`
  * added a hook to simplify using item store state
* ``

### `re-theme`
`[semver-status]: minor`

* `repos/re-theme/src/hooks/useStylesMemo.js`
  * added a hook to simplify memoized joining of theme styles and custom styles

### setup
* to be able to test jsutils, cli, and retheme: `keg hub && keg pr 32`

### `jsutils`
* `keg jsutils`
* `yarn test`
* ensure all tests pass

### `keg-cli`
* start up a tap 
* then run `keg evf sync jsutils:install:start`
* verify it is synchronized into /keg/tap/node_modules/keg-core/@keg-hub/jsutils with `keg sy`
* verify that, when you make a change in jsutils, it rebuilds (you should be able to see the terminal output when you ran the sync command)

### `keg-components`

* `keg d package run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-components:zen-278-keg-component-updates`

#### Checkbox
* open the `checkbox` stories
* in `disabled`, verify all the checkboxes have disabled styling: a `not-allowed` cursor and dimmed opacity
* in `disableCheck`, verify you can only check two boxes at most
* in `disableUncheck`, verify that you can **only** check a box, but not uncheck it

#### CheckGroup
* open the `CheckGroup` stories under form
* try each story, verifying it behaves as you would expect

### `keg-core`
* the new hooks will be tested in the tap's PR
* but you can run this package to verify the test app behaves normally and to verify the tests pass: 
  * `keg d package run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-core:zen-278-keg-component-updates`
  * `keg d package run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-core:zen-278-keg-component-updates command=test`

### `re-theme`
* the new `useStylesMemo` will be tested in the tap's PR
* verify the existing tests pass:
  * keg retheme
  * yarn test


repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -111,18 +121,44 @@ export const CheckboxWrapper = props => {
...
+    toggleIsEnabled ? defaultStyles : disabledStyles
+  )
+
   const typeClassName = useThemeTypeAsClass(
     elThemePath || type,
     'keg-checkbox',
     className
   )
 
+  const defaultPressHandler = setCheckedValue(
```
lancetipton, line 148: This all works but it's a bit confusing. Honestly having a hard time following it. 
May be add some notes?

lancetipton (changes requested):
Just a few small things. Everything work as expected, but it's a bit confusing how

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -111,18 +121,44 @@ export const CheckboxWrapper = props => {
     ...elProps
   } = props
 
+  !RightComponent && console.log({ disableCheck, disableUncheck })
```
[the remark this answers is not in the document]
mpcarolin: removed

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -111,18 +121,44 @@ export const CheckboxWrapper = props => {
...
+    onChange || onValueChange,
+    {
+      enableCheck: !disableCheck,
+      enableUncheck: !disableUncheck,
+    }
+  )
+
+  const pressHandler = toggleIsEnabled
+    ? getOnChangeHandler(isWeb, defaultPressHandler)
+    : () => {}
```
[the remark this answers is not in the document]
mpcarolin: actually didn't even need either; `undefined` does the trick too

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
mpcarolin: I refactored this some and added more inline documentation

daniel112 (commented):
testing all passed.
i left some comments on the jsutils update and checkGroup

repos/jsutils/src/array/omitElement.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Returns a new array containing all the elements of `array`, except for `element`
+ * @function
+ * @param {Array<*>} array
+ * @param {*} element
+ * @param {Function?} selector - optional fn returning a property of an element to compare for omission
+ * @returns {Array} subset of array without element
+ */
+export const omitElement = (array, element, selector=identity) => {
```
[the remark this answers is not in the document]
mpcarolin: It's really just a small shortcut:
```js
const arr = [1,2,3,4,5]
const result = omitElement(arr, 4)
result === [ 1, 2, 3, 5 ]
```
vs.
```js
```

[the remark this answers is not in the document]
mpcarolin: I mainly did it because I started to get into a pattern of passing around a selector function in the code surrounding the set hooks (like `useSet`), and I thought:

```js
omitElement(arr, element, selector)
```

was simpler than:

```js
arr.filter(x => selector(x) !== selector(element))
```

But now that you point this out, I see that it's really not much of a shortcut. I can remove it if that's preferred.

repos/keg-components/src/components/form/checkGroup/checkGroup.stories.js
```diff
@@ -0,0 +1,65 @@
+import React, { useState } from 'react'
```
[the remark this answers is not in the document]
mpcarolin: Can do! I honestly forgot we had set that up.

repos/keg-components/src/components/form/checkGroup/checkGroup.js
```diff
@@ -0,0 +1,115 @@
...
+    : children
+
+  // callback that manages select-all behavior on click of the header checkbox
+  const headerCheckHandler = useCallback(checked => {
+    onGroupPress?.(checked)
+    checkedSetters.map(fn => isFunc(fn) && fn(checked))
+  }, [ checkedSetters ])
+
+  return (
+    <View className={className} style={groupStyles?.main}>
```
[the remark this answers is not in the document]
mpcarolin: That makes sense. Will do

repos/jsutils/src/array/omitElement.js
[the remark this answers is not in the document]
daniel112: gotcha. Personally, I would just use the filter prototype even if this were available for me to use. Since the shortcut doesnt really seem to give any other benefit besides slightly less letters to type.  with the addition of it being an extra dependency + us having to maintain it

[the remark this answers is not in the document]
mpcarolin: yeah agreed. I'll remove it 👍

repos/keg-components/src/components/form/checkGroup/checkGroup.stories.js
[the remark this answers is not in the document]
lancetipton: @mpcarolin - Check out this [PR](https://github.com/simpleviewinc/keg-hub/pull/41) for the Drawer components. It's a better implementation of using MDS. That's the pattern we should follow

repos/re-theme/src/hooks/useStylesMemo.js
```diff
@@ -0,0 +1,28 @@
...
+ * @example
+ * const textStyles = useStylesMemo('todos.text', textStylesObject)
+ * const myStyles = useStylesMemo(
+ *  'settings.button',
+ *  { main: { margin: 5 }},
+ *  inputStyles,
+ *  'someOtherPath.button'
+ * )
+ */
+export const useStylesMemo = (...styleArgs) => {
```
lancetipton, line 22: What's the difference between this hook and the `useStyle` hook in keg-components?
Look at => `keg-components/src/hooks/useStyle`

repos/keg-components/src/components/form/checkGroup/checkGroup.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {*} props.children - components in the group (@see CheckboxGroup.Item)
+ * @param {Object?} props.styles
+ * @param {Object?} props.styles.main
+ * @param {Object?} props.styles.header
+ */
+export const CheckGroup = ({ className, title, children, styles, initChecked=false, onGroupPress }) => {
+  const groupStyles = useStylesMemo('form.checkGroup', styles)
+
+  // if children is defined as a function, it's assumed it needs the groupProps for select-all behavior
+  const useCheckboxHeader = isFunc(children)
```
lancetipton, line 74: The name makes me think this is a `hook`. I hate that this is the case because I like the convention, but React kind of stole it with hooks.
Maybe update it to `renderCheckboxHeader` || `applyCheckboxHeader`

repos/keg-components/src/components/form/checkbox/checkbox.js
```diff
@@ -45,45 +45,61 @@ const Input = StyleInjector(KegInput, {
...
-      }
-    }, [ checkBoxStyles, styles ])
+const Element = React.forwardRef((props, ref) => {
+  const {
+    className,
+    elProps,
+    styles = noPropObj,
+    CheckIcon = Check,
+    checked,
+    ...remainingProps
```
lancetipton, line 55: I like calling these `attrs` for `attributes`. Cause really that's what they are

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -35,19 +35,27 @@ const useCheckedState = (isChecked, themeStyles) => {
...
  * @param {function} onChange - Callback to call when the state changes
+ * @param {boolean} options.disableCheck - `setChecked` can only be called with `true` if this value is false
+ * @param {boolean} options.disableUncheck - `setChecked` can only be called with `false` if this value is also false
  *
  * @returns {function} - The checked state update function
  */
-const setCheckedValue = (isChecked, setChecked, onChange) => {
-  return event => {
-    setChecked(!isChecked)
+const getCheckboxPressHandler = (
```
lancetipton, line 48: This should be called `useCheckboxPressHandler` cause it's a hook

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -86,22 +101,50 @@ const ChildrenComponent = ({ children, className }) => (
...
+    initChecked,
     checked,
     children,
     elType,
     Element,
+    CheckIcon,
     disabled,
+    disableCheck = false,
+    disableUncheck = false,
+    allowAdjacentPress = true,
```
lancetipton, line 141: This is excellent. Didn't know I needed it until you added it.

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -86,22 +101,50 @@ const ChildrenComponent = ({ children, className }) => (
...
     isWeb,
     LeftComponent,
     close,
     onChange,
     onValueChange,
-    ref,
+    setCheckedSetter,
```
lancetipton, line 147: This prop name seems off. I get what it's doing but I think it could be better.
The set part does not make sense, cause really it's just calling a function. It's not setting anything with an `=`
Would be better to use a `ref` and set the ref which follow the rest of the react paradigm. For example =>

```js
// Consumer calls child checkbox
const ConsumerComp = props => {
  const toggleCheckedRef = userRef(null)
  <Checkbox toggleCheckedRef={ toggleCheckedRef } /> 
}

// Then in the wrapper set the ref in the useEffect hook
const CheckboxWrapper = props => {
  // ...other codes

  useEffect(() => {
    toggleCheckedRef && ( toggleCheckedRef.current = setChecked )
  }, [ toggleCheckedRef, setChecked ])
}

```

Or event, just set it inline with out the hook, since we have access to it => 
```js
const CheckboxWrapper = props => {
  // ...other codes
  toggleCheckedRef && ( toggleCheckedRef.current = setChecked )
}
```

repos/keg-core/core/base/hooks/state/useExternalSet.js
```diff
@@ -0,0 +1,126 @@
...
+ * usersSet.delete({ id: '4' }) // calls setUsersInStore with an array that omits user id 4
+ */
+export const useExternalSet = (arr, setArr, selector=identity) => {
+  // checks if arr contains the element, using the comparisonFn if defined
+  const contains = useCallback(
+    (arr, element) => isMember(arr, element, selector),
+    [ selector ],
+  )
+
+  return useMemo(
```
lancetipton, line 53: This is dope but why define it inline?
* Every time the hook is called, all these set methods have to be recreated.
* It seems like you could define it outside of the useMemo hook, and then reuse it?

But maybe I'm missing something

repos/keg-core/core/base/hooks/store/useStoreItems.js
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Turns a path into a camel-cased string, if there is more than one
+ * step in the path. If there isn't, just returns path.
+ * @param {str} path 
+ * @example
+ * makeKeyFromPath('settings.agendaMap') -> 'settingsAgendaMap'
+ * makeKeyFrompath('settings') -> 'settings'
+ */
+const camelCasedPath = (path) => {
```
lancetipton, line 64: This seems like it should go into `jsutils`

repos/keg-core/core/base/hooks/store/useStoreItems.js
```diff
@@ -0,0 +1,75 @@
...
+  )
+}
+
+/**
+ * selectPath: gets the path from items, so long as both exist and path is a string
+ * @param {object} items 
+ * @param {string} path 
+ * @return {*?} - the value stored in items at path
+ */
+const selectPath = (items, path) => {
```
lancetipton, line 32: This method seems kind of point less. If you need this, update jsutils `updateColl` method to check it the path is an array or string.
Then on line 49 just call `get` like this => `get(items, path, null)`

repos/re-theme/src/hooks/index.js
```diff
@@ -3,3 +3,4 @@ export * from './useStylesCallback'
 export * from './useThemeHover'
 export * from './useThemeActive'
 export * from './useThemeFocus'
+export * from './useStylesMemo'
```
lancetipton, line 6: This seems to be a duplicate of the `useStyle` hook is keg-components.

repos/re-theme/src/hooks/useStylesMemo.js
```diff
@@ -0,0 +1,28 @@
+/** @module hooks */
+
+import { useMemo } from 'react'
+import { useTheme } from './useTheme'
+import { deepFreeze } from '@keg-hub/jsutils'
+
+const noOpObj = deepFreeze({ content: {} })
```
lancetipton, line 7: * Update this to pull from `../helpers/noOp.js`
* Then in `../helpers/noOp.js` import the `noOpObj` from jsutils

lancetipton (changes requested):
Looking really good. Lots of really good updates / changes from last time. Much cleaner.
Just a few things too look at.

repos/re-theme/src/hooks/useStylesMemo.js
mpcarolin: I forgot that hook existed. 

This one, `useStylesMemo`, originates from that call we all had a couple of weeks ago  about this topic.

There are two differences I see:
* if `theme.get(...)` returns `undefined`, `useStylesMemo` returns the `noOpObj` that has a `content` key instead. `useStyles` will just return `undefined`. This saves the consumer from needing to null-check the return value.
* `useStyles` recomputes the styles only if one or more of the arguments changed. `useStylesMemo` recomputes it if either the arguments change OR the retheme theme object itself changes. One motivation for this was to recompute when the current window size changes. I don't think `useStyles` would necessarily recompute in that case. Although, with the recent css style changes, that may not be needed?

The only other difference is that it's in `re-theme`, which to me is more appropriate.

What are your thoughts? Now that I see `useStyles` exists, my preference is to delete `useStyles` from re-theme and then rename `useStylesMemo` to `useStyles`.

mpcarolin: Talked on slack: going to move ahead with deleting `useStyles` from keg-components and then renaming `useStylesMemo` to `useStyles`

repos/keg-core/core/base/hooks/state/useExternalSet.js
mpcarolin: Because they each depend on the `arr` argument, and so if they were defined separately, they would either get out of date quickly or would need to be recreated separately.

I _did_ consider using a class, something like:

```js
class ExternalSet {
  constructor(array, setArray) {
    this.array = array
    this.setArrray = setArray
  }

  add (element) {
    pipeline([ ...arr, element ], uniqArr, setArr))
  }
 
   ...
}
```

and the hook would be something like:

```js
const useExternalSet = (arr, setArr, selector) => {
  return useMemo(() => new ExternalSet(arr, setArr, selector), [ arr, setArr, selector ])
}
```

Of course, now it needs to instantiate a new `ExternalSet` class instance every time instead of just an object literal with functions. I'm not sure that would be any faster though (and historically I've seen classes be really slow). On the other hand, it might not recreate the functions, since it could just reuse the ones defined on the prototype.

In the end, I thought any performance differences with this would be minor, but using the class would be more complex, so I chose the object instead.

lancetipton: wow, nice explanation. Seems like you really spent time thinking about it. That works for me.

repos/jsutils/src/array/omitElement.js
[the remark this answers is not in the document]
mpcarolin: updated

repos/keg-components/src/components/form/checkGroup/checkGroup.stories.js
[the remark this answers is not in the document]
mpcarolin: updated

repos/keg-components/src/components/form/checkGroup/checkGroup.js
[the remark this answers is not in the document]
mpcarolin: updated

repos/keg-components/src/components/form/checkGroup/checkGroup.stories.js
```diff
@@ -0,0 +1,65 @@
...
+      <StoryWrap>
+        <CheckGroup title={'Goats'} initChecked={initialCheckValue} check>
+          { ({ setCheckedSetter }) =>
+              goats.map(breed => {
+                const [ value, setValue ] = useState(initialCheckValue)
+                React.useEffect(() => void setCheckedSetter(setValue), [ setValue ])
+                return (
+                  <div key={breed} style={{ display: 'flex', flexDirection: 'row'}}>
+                    <input type='radio' checked={value} onChange={setValue}></input>
+                    <p>{ breed }</p>
```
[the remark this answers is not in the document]
mpcarolin: updated

Also changed it to use switch, and updated switch also work with checkgroup (accepts a ref)

repos/jsutils/src/method/noOp.js
```diff
@@ -0,0 +1,8 @@
+/** @module functions */
+
+/**
+ * Just a reusable empty function that is a no-op
+ * @function
+ * @returns {void}
+ */
+export const noOp = () => {}
```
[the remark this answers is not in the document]
mpcarolin: updated

repos/re-theme/src/hooks/useStylesMemo.js
mpcarolin: updated

repos/keg-components/src/components/form/checkGroup/checkGroup.js
mpcarolin: good point. updated

repos/keg-components/src/components/form/checkbox/checkbox.js
mpcarolin: updated

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
mpcarolin: Thanks, updated

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
mpcarolin: 👍

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
mpcarolin: Yeah, I agree it was a bit confusing.

I've changed the checkbox and the group to work with refs.

I decided to use the `useImperativeHandle` hook, to define the api the consumer can use (in this case, `setChecked` or `isChecked`). I prefer that because we expose only what the consumer needs, it uses the standard ref argument of react, and I think it's  makes clear what is going on.

 Let me know what you think.

repos/keg-core/core/base/hooks/store/useStoreItems.js
mpcarolin: yep, you're right, updated! We use paths everywhere so hopefully this will come in handy elsewhere.

repos/keg-core/core/base/hooks/store/useStoreItems.js
mpcarolin: agreed, removed. It can just be a plain get call since it returns defined when path is empty/null.

repos/re-theme/src/hooks/index.js
mpcarolin: spoke on slack -- updated to delete the `keg-components` version and renamed this hook to be `useStyle`

repos/re-theme/src/hooks/useStylesMemo.js
mpcarolin: updated

repos/keg-components/src/components/form/checkGroup/checkGroup.stories.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+<Preview>
+  <Story name='With any checkable component'>
+    <WithSwitch />
+  </Story>
+</Preview>
+
+### Props
+
+<Props of={CheckGroup} />
```
daniel112, line 46: the prop data isn't showing up for me 
image

daniel112 (changes requested):
looks good. just 1 last tiny thing

lancetipton (changes requested):
Looks good. Left a couple comments on things to be looked at

repos/keg-components/src/components/form/checkGroup/checkGroup.js
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Simple header for CheckGroup, without a checkbox
+ * @param {Object} props
+ * @param {string} props.title - title of header
+ * @param {Object} props.style - style rules to apply to the Text element
+ * @param {*} props.* - remaining props are passed to the Text element
+ */
+export const SimpleHeader = ({ title, style, ...rest }) => {
+  const textStyle = useMemo(() => ({ ...style, marginLeft: 27 }), [ style ])
```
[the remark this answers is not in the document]
mpcarolin: updated to be able to be overwritten by `style` prop, and margin is coming from the theme file now.

repos/keg-components/src/components/form/checkbox/checkbox.js
```diff
@@ -45,53 +44,70 @@ const Input = StyleInjector(KegInput, {
...
+    elProps,
+    styles = noPropObj,
+    CheckIcon = Check,
+    checked,
+    ...attributes
+  } = props
+
+  const checkStyle = useMemo(() => {
+    return {
+      ...checkBoxStyles.icon,
```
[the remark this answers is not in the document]
mpcarolin: Since I'm using a custom icon, I needed to style it differently (mainly width and height) to match the events force mock up.

But why should they not be overwritten? I get that in 99% of cases, the styles need to stay this way otherwise things look wrong. But haven't we always said that we want the styles to be fully configurable, because of how frustrating it is when other component libraries don't allow that?

I understand why the `input` styles should be locked, because the input isn't actually visible at all and exists solely for the functionality of the component. I'll definitely revert that change (I'm not overwriting the input styles).
 
But the icon is visible and I think it makes sense to let consumers style it if they need to, no?

repos/keg-core/core/base/utils/events/validateEventResponse.js
```diff
@@ -0,0 +1,18 @@
...
+ * The keg event emitter returns a boolean value when calling `.emit(...)`, indicating
+ * whether or not a callback was registered against the event and run.
+ *
+ * This is a helper to handle that response.
+ * @param {boolean} isValid - the response to `.emit`
+ * @param {Array<string>} invalidMessages - messages to log if the result was invalid
+ * @param {Array<string>} valideMessages - messages to log if the event was valid
+ * @return {boolean} - the value of isValid
+ */
+export const validateEventResponse = (
```
daniel112, line 11: is this made to wrap
```Javascript
  const valid = kegEventEmitter.emit(
    EVENTS.SESSION_BOOKING_REQUEST,
    data
  )
  if (!valid)
    console.warn(
      `Callback for ${EVENTS.SESSION_BOOKING_REQUEST} does not exist!`
    )
```

to this
```Javascript
  validateEventResponse(kegEventEmitter.emit(
    EVENTS.SESSION_BOOKING_REQUEST,
    data),
   ['some success messages'],
   ['some error messages']
  )
```

daniel112: resolved in chat

daniel112 (approved):
works for me

repos/keg-components/src/components/form/checkGroup/checkGroup.js
```diff
@@ -0,0 +1,139 @@
...
+          checked={check ? initChecked : undefined}
+        />
+      ) : (
+        <SimpleHeader
+          className={headerClassName}
+          style={groupStyles?.header}
+          title={title}
+        />
+      ) }
+      { /* { renderedChildren } */ }
```
lancetipton, line 125: This comment seems pointless

repos/keg-components/src/components/form/checkGroup/checkGroup.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * A group of checkbox items with a header.
+ * Will include a select-all checkbox in the header IF you define the children as a function (@see story examples)
+ * @param {Object} props
+ * @param {string} props.className - css class name
+ * @param {string} props.headerClassName - css class name for header text
+ * @param {string} props.title - title of the group
+ * @param {boolean?} props.initChecked - initial checked value of the group header box, if you are using it
+ * @param {boolean?} props.check - true if you want a group header check box
```
lancetipton, line 72: This prop name seems odd `check` to me does not indicate it shows the GroupHeader checkBox
Also I would add a way to disable the header completely.
There may be cases where I don't want it, but I want the checkbox group

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -82,27 +107,84 @@ const ChildrenComponent = ({ children, className }) => (
...
+ * @param {string} props.type - theme type
+ * @param {string} props.themePath - path to theme object
+ * @param {boolean} props.value - alias for initChecked
+ * @param {...rest} props.* - the remaining props are passed to Element
  */
-export const CheckboxWrapper = props => {
+export const CheckboxWrapper = forwardRef((props, ref) => {
   const {
     className,
+    initChecked,
```
lancetipton, line 169: lol... that's a lot of props just for a checkbox

repos/keg-components/src/components/form/checkGroup/checkGroup.js
```diff
@@ -0,0 +1,139 @@
...
+ * @param {Function?} props.onGroupPress - handler of header checkbox
+ * @param {*} props.children - components in the group (@see CheckboxGroup.Item)
+ * @param {Object?} props.styles
+ * @param {Object?} props.styles.main
+ * @param {Object?} props.styles.header
+ */
+export const CheckGroup = props => {
+  const {
```
lancetipton, line 80: We should add a prop for getting access to the `headerCheckHandler`
* This way the parent can get access to the method an call it externally
  * In the `checkbox.example.js`, you give an example of checking a single checkbox externally, but there's not way to check all checkboxes externally.
  * You could use a ref again, and set the current to be `headerCheckHandler` or something to that effect
  * Just try to keep it consistent across the components

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -82,27 +107,84 @@ const ChildrenComponent = ({ children, className }) => (
...
+    disableUncheck = false,
+    allowAdjacentPress = true,
     isWeb,
     LeftComponent,
+    leftClassName,
     close,
     onChange,
     onValueChange,
-    ref,
+    setCheckedSetter,
```
lancetipton, line 185: This is no longer being used and should be removed

repos/keg-components/src/utils/helpers/handleRefUpdate.js
```diff
@@ -8,6 +8,9 @@ import { isObj, checkCall } from '@keg-hub/jsutils'
  * @returns {Object} - Combined child components
  */
 export const handleRefUpdate = (ref, update) => {
+  if (!ref) return update
```
lancetipton, line 11: These 2 lines, could just be this => `if(!ref || !update) return ref || update`
Although, I'm not sure the code is correct.
* If `ref` is a valid ref, and `update` is `false`, It should still update the `ref.current` with `false`
* But with this line `if (!update) return ref`, that never happens

So I think something like this would be better =>
```js
export const handleRefUpdate = (ref, update) => {
  // Update the ref based on it's type
  return !ref
    ? update
    ? isObj(ref) && 'current' in ref
      ? (ref.current = update)
      : checkCall(ref, update)
}
```

Might need some testing to ensure I didn't miss something.

lancetipton (changes requested):
Getting closer. Just a few things to look at.

repos/keg-components/src/components/form/checkGroup/checkGroup.js
mpcarolin: updated

repos/keg-components/src/components/form/checkGroup/checkGroup.js
mpcarolin: updated 
- changed prop name to `showHeaderCheckbox`
- added prop `showHeader`, true by default, but hides the header completely when false

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
mpcarolin: Yeah, it's doing a lot, but on the other hand, it's super versatile

repos/keg-components/src/components/form/checkGroup/checkGroup.js
mpcarolin: Agreed. Updated `Checkgroup` to accept a ref, which will provide access to the parent's ref. Also added a story to demonstrate select-all using this ref.

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
mpcarolin: updated

repos/keg-components/src/utils/helpers/handleRefUpdate.js
mpcarolin: tested and updated 👍

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
```diff
@@ -82,27 +107,84 @@ const ChildrenComponent = ({ children, className }) => (
...
+ * @param {boolean} props.disableCheck - disables checking if true
+ * @param {boolean} props.disableUncheck - disables unchecking if true
+ * @param {boolean} props.allowAdjacentPress - if true, allows pressing a side component to toggle the checkbox
+ * @param {boolean} props.isWeb - indicates current platform is web. Provide by checkbox.js
+ * @param {(Component | string)?} props.LeftComponent - an optional component or string to render left of the checkbox
+ * @param {string?} props.leftClassName - optional class name for the left component
+ * @param {boolean} props.close - if true, side components will be rendered close to the checkbox. If false, spaced apart.
+ * @param {Function?} props.onChange - callback fired when checkbox is toggled
+ * @param {Function?} props.onValueChange - callback fired when checkbox is toggled
+ * @param {Function?} props.setCheckedSetter - callback called once on initial render. Is passed the `setChecked` callback to provide consumer control of checked/unchecked state if needed.
```
lancetipton, line 156: This prop no longer exists. You should remove the notes too.

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
lancetipton: It's removed from the props, but it still exists in the comments. Can you remove it from line 156 as well

lancetipton (approved):
Only needs one small thing and should be good. I'm going to approve because it's just a comment that needs deleted. Everything else looks good.

repos/keg-components/src/components/form/checkbox/checkbox.wrapper.js
mpcarolin: yep updated

PR #36, opened 2020-09-21
keg-components: scrollView wrapper

daniel112:
[Semver Status] minor

## Context

* we're using `ScrollView` from `react-native` in the tap
* Currently you cant pass in `className` to `ScrollView` for react-native-web, but maybe they might find a use for it and update it in the future
* lets just wrap the `ScrollView` so its import is consistent in the tap

## Goal

* wrapping `ScrollView` in keg-components

## Updates

* `repos/keg-components/src/components/native/index.js`
   * exporting `ScrollView` from `react-native`

## Testing
see testing here https://github.com/simpleviewinc/tap-events-force/pull/56


lancetipton:
Closing, this is being handled in #38

PR #38, opened 2020-09-23
Zen 355 style injector hook

lancetipton:
**Ticket [ZEN-355](https://jira.simpleviewtools.com/browse/ZEN-355)** 

### Updates - 9/24
**Fixes**
* Applied fixes for all issues brought up in the PR
* Pushed new docker packages for Tap and Components 

**Added**
* Added important rules
  * Get `!important` added to them to override RN-Web defaults
  * This allows Touchable transition-duration to work properly
    * Here's an [example of what it looks like](http://snpy.in/Yf3fif)
    * This is mostly to fix/override [this](https://github.com/necolas/react-native-web/blob/af0d80a808f6d246facc38fac1cd24d2b205f0ac/packages/react-native-web/src/exports/TouchableOpacity/index.js#L136)
      * The `duration` value gets set on [line 91](https://github.com/necolas/react-native-web/blob/af0d80a808f6d246facc38fac1cd24d2b205f0ac/packages/react-native-web/src/exports/TouchableOpacity/index.js#L91) to be either `0 || 150`
      * Using `!important` allows us to override that value with our own value
      * It's only used when passed in through the style prop
* Moved default filter and important rules to their own constants file
* Fixed a bug with custom filter rules

<br/>
<br/>

**---- END UPDATES ---**

<br/>
<br/>

> **Notes**
> This PR is directly tied to [this PR](https://github.com/simpleviewinc/tap-events-force/pull/57)
>   * They should be merged at the same time
>
> Originally this ticket was to add a `react-helmet`
>   * I could not get that library to not de-dupe styles that were already added
>   * I knew it had to be possible, so I wrote it myself
>
> Because of this, the jira ticket and this pull request don't really line up
>   * This should be the last styling change ticket. At least I hope it is
>   * I'm tired of dealing with it

## Context
* Events-Force is defining their styles in CSS
* Their clients can define custom CSS as well
* Currently, the tap, using re-theme, is only compatible with JSON style objects
* We need to be able to pull styles from stylesheets already applied to the DOM and apply them to keg-components

## Goal
* Use the already added className prop, to allow applying styles useing the dom
  * This should only happend when in a web context
  * Native context, the style attribute should still be used
  * Ensure no changes need to be added to taps when defining styles
  * Ensure styles are not duplicated on the Dom
    * This is gold right here. I can't find any other repo that does this
    * It puts ReTheme on another level in my opinion 
* Taps can now do the following
  * Use `.css` styles applied to the dom and classNames **(web only)**
  * Use the style attribute, and still get all the benefits of the Dom
    * This works cross-platform

## Updates
**ReTheme**
* `Global Updates`
  * Added more tests
    * Coverage is still only at `78%`, but with the added code, at least it did not go down
    * Pretty sure I have full coverage on all newly added code
    * Also added tests for a number of files that had none
  * Removed a number of files / code no longer being used
* `repos/re-theme/configs/rollup.config.js`
  * Update the exports
  * Remove the head export
  * Added styleInjector exports
  * Added the `ReDimensions` alias back
    * Fixes issues on mobile, so it is still needed
* `repos/re-theme/package.json`
  * Added alias to the jest config for `ReDimensions`
    * Resolves to web by default
* `repos/re-theme/src/constants/constants.js`
  * Removed a bunch of unused constants
  * Added a `KEG_STYLES_TAG_ID` constant
* `repos/re-theme/src/dimensions/dimensions.native.js`
  * Added to only be used on native
  * Will auto-load `react-native Dimensions` export if it's not set
  * Resolves issues with a race condition, where it was called to early, and causes app crashes
* `repos/re-theme/src/head/*`
  * Remove all this code. No longer needed, and I'm very happy about that
* `repos/re-theme/src/helpers/colors.js`
  * Fixed a few bugs I found while writing tests
* `repos/re-theme/src/styleInjector/reactNativeWeb.js`
  * Put all `react-native-web` imports in the same file
  * This way if they need to be updated, it's in one location
* `repos/re-theme/src/styleInjector/styleInjector.js`
  * A `HOC` that wraps a passed in component, and extracts the styles from the props
  * It then converts those styles to valid Css Dom styles and appends them to the Dom
    * Uses the `useStyleTag` hook
  * Keeps track of added styles, and only adds the styles if they have not already been added
* `repos/re-theme/src/styleInjector/useStyleTag.js`
  * This is the core code that allows classNames / Dom Css styles to work
  * Uses some `react-native-web` exports to do the style conversion
    * **This allows for consistent styling between Css styles and the style attribute**
* `repos/re-theme/src/styleParser/styleSheetParser.js`
  * Fixed issues with the styleSheetParser
  * Now passes in a cssToJs helper to the callback
    * This way we don't have to `re-write / import` it every time it's used
    * But hopfully we'll never have to do this again
  * Removed the default callback. A callback is now required as an argument
* `repos/re-theme/src/styleParser/validate.js`
  * Updated to validate the `styleSheetParser` changes
* `repos/re-theme/src/theme/restructureTheme.js`
  * Fixed a bug that was causing the correct platform styles to not always load properly
    * This happened in chases where the a `$<platform>` exists, but does not match the current platform
      * In this case if would return the original theme object with the `$<platform>` still part of the theme
      * Pretty sure this is a bug @daniel112 was running into about a year ago
      * I noticed it while getting the tap to run on mobile

**Keg-Components**
* `Global Updates`
  * Updated more the the primitives to use the `StyleInjector`
    * This allows adding the stlyes to the Dom, and is tied to the component with a className
    * The general pattern I tried to follow looks like this =>
      * `< component >.js`
        * Standard web export
        * Imports the `.native.js` version of the component
        * Wraps the native component with the `StyleInjector`
      * `< component >.native.js`
        * Standard native export
          * Does not call the `StyleInjector` in most cases
          * `StyleInjector` is stubbed out on native, so it's not an issue even it it's called
* `repos/keg-components/configs/rollup.config.js`
  * Minor clean up
* `repos/keg-components/src/components/**`
  * Updated to use the `StyleInjector` Hoc, and `useStyleTag` hook from ReTheme
* `repos/keg-components/configs/aliases.json`
  * Cleaned up alias based on changes
* `repos/keg-components/src/components/header/appHeader.js`
  * Added an `appHeader` prop that gets passed to the `itemHeader`
    * If this is true, then it will add the default appHeader styles from the theme
* `repos/keg-components/src/components/header/itemHeader.js`
  * Added the `appHeader` prop
* `repos/keg-components/src/components/image/image.js`
  * Updated to use the `KegInternalImage` component
    * Updated how styles and classNames are applied to the image
    * This update is cleaner, and seems to have better performances
* `repos/keg-components/src/components/list/scrollView.js`
  * Wrapped the scrollView component from react-native
  * Added StyleInjector to allow passing className
* `repos/keg-components/src/components/touchable/touchable.js`
  * Minor cleanup
* `repos/keg-components/src/hooks/useClassName.js`
  * Added tracking of the passed in `classNames`
    * Compares the **new classNames**, with the **old classNames**
    * Adds any new classes, and removes any old classes
* `repos/keg-components/src/hooks/useFromToAnimation.js`
  * Added `useNativeDriver` when not on a web platform
* `repos/keg-components/src/hooks/useTextAccessibility.js`
  * New hook only used in the KegText component
  * Figures out the `accessibilityRole` for the component
* `repos/keg-components/src/hooks/useTextStyles.js`
  * New hook to get the default styles for typography components
* `repos/keg-components/src/theme/components/header/itemHeader.js`
  * Minor cleanup to the `shadow` styles for Item header

**NOTES**
* `Animated.*` styles are not added to the Dom
  * If should be possible, but is very complicated, I was not able to get it to work properly
    * A lot of the animations use javascript to do the work
    * I'd have to figure out a way to match the javascript animations with the CSS animations
    * I can see how it would be possible, but I don't think it's worth the time
      * Especially because they work just fine right now, so why bother

**Keg-Core**
  * Fixed issue with running on mobile device
  * Added `jsutils` to resolved alias
    * Ensure all repos use the same jsutils
  * 

**Keg-CLI**
* `repos/keg-cli/containers/components/mutagen.yml`
  * Added attach actions, to make it easier to navigate to sub-repos in node_modules folders
    * Example => `keg evf sync retheme:att` === Attaches to the docker container, and navigate to the re-theme directory

## Testing

**Tap**
> **Notes**
> This is the same tests as [this PR](https://github.com/simpleviewinc/keg-hub/pull/38)
> So if you already did the tests for that one, they you have test this one

* Run Docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-355-styles-fixes`
  * Ensure it runs
  * Open the inspector and check out the `Dom Tree`
    * Styles should no longer be applied to the elements with the style attribute
      * [All View / Text components should be using a class](http://snpy.in/MwwBpH)
      * [Some will still have styles ( Animated elements / Some primitives )](http://snpy.in/TuQOGY)
      * [Some will have classes and styles for animations (buttons)](http://snpy.in/3d2mAV)
* Test Docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-355-styles-fixes test`
  * Ensure all tests pass

**Keg-Components**
* Run Docker package => `keg dpg run docker.pkg.github.com/simpleviewinc/keg-packages/keg-components:zen-355-styleinjector-hook`
  * Ensure all components display properly
  * Styles should work the same as the Tap
    * With classNames and some with styles for animations
    * Open the inspector and check out the `Dom Tree`
      * Ensure classNames are being applied properly
      * Ensure all components render properly
* Test docker package => `keg dpg run docker.pkg.github.com/simpleviewinc/keg-packages/keg-components:zen-355-styleinjector-hook test`
  * Ensure all tests pass

**ReTheme**
* No package to run because you can test the changes in Keg-Components, and the Tap packages above
* Run ReTheme tests
  * `keg retheme && keg pr 38 && yarn install && yarn test`
  * Ensure all tests pass


repos/keg-components/src/utils/helpers/updateClassNames.js
```diff
@@ -0,0 +1,34 @@
+import { getPlatform } from 'KegGetPlatform'
+import { ensureClassArray } from '../../utils/helpers/ensureClassArray'
+const isWeb = getPlatform() === 'web'
+
+/**
+ * Helper to update the classNames on a Dom node
+ * @param {Object} element - Dom node element
+ * @param {Array[string]} classesRef - Array of classes currently on the element
```
mpcarolin, line 8: I think for maximum clarity, the type of `classesRef` should be 
`RefObject<Array<string>>` (replace `< and >` with `[ and ]` if you wish)

repos/re-theme/src/styleInjector/injectHelpers.js
```diff
@@ -0,0 +1,222 @@
...
+ * <br/>Value comes from being a pure function
+ * <br/>Given the same input, it will always return the same output
+ * <br/>There is no expectation to convert back from the hash to the original string
+ * @function
+ * @param {string} str - String to be hashed
+ * @param {number=} maxLength - Max length of the returned hash
+ *
+ * @returns {string} - Hashed version of the string
+ */
+export const hashString = (str, maxLength=0) => {
```
mpcarolin, line 88: why not in `jsutils`?

repos/re-theme/src/styleInjector/injectHelpers.js
```diff
@@ -0,0 +1,222 @@
...
+ */
+export const getSelector = (className, cssString) => {
+  const selector = !exists(className)
+    ? false
+    : isArr(className)
+      ? className.filter(cls => cls).join('.').trim()
+      : isStr(className) && className.split(' ').join('.').trim()
+
+  return selector
+    ? `.${selector.trim()}.keg-${hashString(cssString)}`.trim()
```
mpcarolin, line 154: I think there are maybe some redundant trim calls, since you trim `selector` here and also on lines `150` and `151`.

repos/re-theme/src/styleInjector/injectHelpers.js
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * Cache holder to hold the main StyleSheet Dom element
+ */
+let KegStyleSheet
+
+/**
+ * Caches selectors already added to the Dom
+ */
+const selectorCache = new Map()
```
mpcarolin, line 21: I think this might be better as a `Set` type, since you're only making existence checks, right?

repos/re-theme/src/styleInjector/injectHelpers.js
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * Converts a camelCase style rule into a hyphenated style rule
+ * <br/>Caches the response to make future conversions faster
+ * @function
+ * @param {string} str - camelCase style rule rule
+ *
+ * @returns {string} - Hyphenated style rule
+ */
+export const hyphenator = rule => {
```
mpcarolin, line 69: image

mpcarolin (changes requested):
Overall this looks great to me. `StyleInjector` is awesome.

I have some comments

1. `input` elements for `checkbox` and our `TextInput` components still had a `style` attribute. I just want to confirm: that's ok?
image

2. One retheme test failed for me (after a full clean too):
image

3. Made a handful of code comments, but nothing major.

daniel112 (changes requested):

lancetipton (commented):
Fixed issues, and pushed new docker packages
* Tap => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:keg-mobile-build`
* Keg-Components => `keg dpg run docker.pkg.github.com/simpleviewinc/keg-packages/keg-components:zen-355-styleinjector-hook`

* Also fixed tests for Retheme. So they should all be passing now.

daniel112 (approved):
looks good to me

repos/keg-components/src/utils/helpers/updateClassNames.js
mpcarolin: yep that makes sense to me 👍

repos/re-theme/src/styleInjector/injectHelpers.js
mpcarolin: all good

mpcarolin (approved):
Looks good 👍

PR #40, opened 2020-09-24
Keg-Components && Keg-Core update version to 6.0.0

lancetipton:
## Context

* Update Keg-Components to version 6.0.0
* Include new build
* [Published Keg-Components to NPM](https://www.npmjs.com/package/@keg-hub/keg-components)
* [Published Keg-Core to NPM](https://www.npmjs.com/package/@keg-hub/keg-core)

### Updates
 * Updated jest config to allow all tests for pass in keg-core
   * Had to add react-native-web to be transpiled when tests run
   * the common.js option was removed from app.json, so it tries to load the esm version
   * For development / building this is better, but for tests it fails
     * Transpiling before running tests fixes the issue, but takes a little longer for the tests to run


daniel112 (approved):
looks good to me.
i verified by pulling your tap PR, installing `keg-core` 6.0.0 and running it locally

PR #45, opened 2020-10-08
added the other noOp obj

daniel112:
**Ticket**: n/a

## Context

* adding more noOp items in keg-core for the Tap to use

## Goal

* adding more noOp items in keg-core for the Tap to use so I can use the `noPropArr` in the tap

## Updates

* `repos/keg-core/core/base/utils/helpers/method/noop.js`
   * added noOpObj, noPropObj, noPropArr

## Testing
* n/a


daniel112:
michael is doing it

PR #47, opened 2020-10-12
keg-components: Text toggle component

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-391)
`[semver]: minor`

## Context

* The Session details on mobile has this expandable text component
    * image

* we currently don't have a component like that
* create a component that can do this:
   * cuts off text given a max # of lines
   * has a toggle (show more/less) button
   * when the toggle button is clicked, show all the text 

## Goal

* create `TextToggle` component that can do those things stated in **Context**
* add the appropriate stories and examples
* utilize `Drawer` component within the `Text Toggle` component
* update `Drawer` component slightly so we can set the collapsed height at any point, as opposed to first render

## Updates

**Keg-Components**
* installed `react-native-linear-gradient` and `react-native-web-linear-gradient`

* `repos/keg-components/src/components/textToggle/textToggle.js`
   * TextToggle component
* `repos/keg-components/src/components/textToggle/textToggle.stories.mdx`
   * stories for textToggle
* `repos/keg-components/src/components/textToggle/textToggle.examples.js`
    * examples for TextToggle that's utilized in the stories.mdx

* `repos/keg-components/configs/babel.config.js`
   * added the dependencies to ensure the props table work - same as the one @mpcarolin used in his PR

* `repos/keg-components/src/components/drawer/drawer.examples.js`
   * no changes, husky formatting

* `repos/keg-components/src/components/drawer/drawer.js`
   * eslint formatting
   * updated `initial` prop to be `collapsedHeight`
   * also updated so we can set the collapsedHeight at any point as opposed to only on first render

**Keg-Core**
* added dev dependency for `react-native-linear-gradient`

## Testing

* Run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-components:text-toggle-component`
* navigate to http://local.kegdev.xyz/?path=/docs/interactions-texttoggle--basic
* verify the toggling of the text works


repos/keg-components/src/components/textToggle/textToggle.js
```diff
@@ -0,0 +1,179 @@
...
+ * @param {string} props.fadeColor - optional color for the fade on collapsed text
+ */
+export const TextToggle = props => {
+  const {
+    text,
+    styles,
+    isExpandedInit = false,
+    className,
+    CustomToggle,
+    onToggleChange,
```
lancetipton, line 56: Why not just `onToggle` ?

repos/keg-components/src/components/textToggle/textToggle.js
```diff
@@ -0,0 +1,179 @@
...
+ * @param {'left'|'center'|'right'=} props.togglePosition - optional. where to position the toggle component. default 'right'
+ * @param {Number=} props.collapsedHeight - optional. height of the textview when collapsed
+ * @param {Component=} props.CustomToggle - optional toggle component to use instead of the default
+ * @param {string} props.fadeColor - optional color for the fade on collapsed text
+ */
+export const TextToggle = props => {
+  const {
+    text,
+    styles,
+    isExpandedInit = false,
```
lancetipton, line 53: This seems overly verbose, just to say `expanded`

repos/keg-components/src/components/textToggle/textToggle.js
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * ToggleComponent
+ * @param {object} props
+ * @param {Function} props.onPress - when the component is pressed
+ * @param {object} props.styles
+ * @param {Component=} props.CustomComponent - optional Component to override the default text component
+ * @param {boolean} props.isExpanded - whether to show expanded state or not
+ */
+const ToggleComponent = ({ onPress, styles, CustomComponent, isExpanded }) => {
```
lancetipton, line 152: We should add props the override the default text. I may not want a custom component. I just want to change what the text is. Something like `closedToggleText` and `expandedToggleText`

lancetipton (approved):
Looks good. Made a few comments but nothing blocking. So make the changes only if you think it's worth it.

repos/keg-components/src/components/textToggle/textToggle.js
daniel112: you right, updated

repos/keg-components/src/components/textToggle/textToggle.js
daniel112: i want to signify that setting that will not change the toggled state after the first render

repos/keg-components/src/components/textToggle/textToggle.js
daniel112: agreed. updated

PR #51, opened 2020-10-19
Clean up prefix logic, and fix docker exec task

lancetipton:
## Context
* There were some issue finding the correct docker image based on the prefix
* There was also some code to work around a bug in args parse, that has since been resolved.
  * So the work around code is no longer needed. [See this commit](https://github.com/simpleviewinc/keg-hub/commit/e19f1a61744ee25ea7d82301a045f9cafd2b3498#diff-d4a21c163185fcbe50b72bd0ddb24478a36717d7014a07823816a7c75e407912)

## Goal
* Fix the prefix issue in the docker exec task
* Normalize prefix usage for all tasks
* Removed args-parse work around

## Updates

* `repos/keg-cli/containers/tap/mutagen.yml`
  * Removed the attached default. Can be passed in as a param if still needed
* `repos/keg-cli/src/tasks/docker/exec.js`
  * Removed custom context code, and updated to use the shared context code
* `repos/keg-cli/src/utils/builders/__tests__/buildCmdContext.js` && `repos/keg-cli/src/utils/getters/__tests__/getContext.js`
  * Updated test to match code changes
* `repos/keg-cli/src/utils/builders/buildContainerContext.js` && `repos/keg-cli/src/utils/docker/getContainerFromContext.js`
  * Updated to use `withPrefix` key
* `repos/keg-cli/src/utils/getters/getContext.js`
  * Removed `args-parse` work around
  * Updated to use `withPrefix` key 
* `repos/keg-cli/src/utils/getters/getPrefixContext.js`
  * Updated to set the `withPrefix` key to the value of what `prefix` used to be
  * The `prefix` key is now set to the actual `prefix` of the docker image
    * This improves clarity on what the value of the key actually is
    * This gives access to the `prefix` without needing to parse it again
    * Moved logic to build the prefix data into it's own method for clairity => `buildPrefixData `

## Testing
* Pull with PR - `keg cli && keg pr 51`
**Test 1**
* Pull the docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-371-booking-button-states`
* Navigate to http://local.kegdev.xyz/
  * Ensure it loads properly in your browser
* Open another terminal window
  * Run the command `keg d exec --context package-tap-zen-371-booking-button-states --cmd \"ls -la\"`
    * It should print out a list of files from the default entrypoint of the container `/keg/tap`
  * Next, run the command `keg d att package-tap-zen-371-booking-button-states`
  * This should attach to the running docker container of the docker-package
    * Enter the command `ls -la`
    * Ensure the output is the same as the previous exec command
    * **Note**
      * The `exec` && `attach` tasks did not work before for `docker package` containers
      * The `attach` task is just the `exec` task with some defaults

**Test 2**
* Start any of the other repos to ensure the commands are working properly
* Run the commands
  * `keg base build`
  * `keg components start`
      * Uses the `exec` task internally
      * Should auto build the components container if it does not exist
      * Should start the components app in the container
  * `keg components att`
    * Uses the `exec` task internally
    * Should attach to the components container
  * `keg components stop`
    * Should stop, but not kill the components container
  * `keg core start --sync components:install:build`
    * Uses the `exec` task in a number of ways when using the `--sync` option
  * `keg core att`
    * Uses the `exec` task internally
  * `keg core kill`
    * Should fully remove the core container
  * `keg evf start --sync core:install`
    * Tests running and syncing taps
    * Uses the `exec` task in a number of ways when using the `--sync` option
  * `keg retheme build`
    * This assumes you have the re-theme repo linked as an injected app
    * Test building injected apps
  * `keg retheme start`
    * Test running injected apps

**Test 3**
* Go back to the original terminal and exit the running container
* Run `keg cli test`
  * This will run the `jest tests` for the `keg-cli` repo
  * Ensure all tests pass
* **Important** - There are some docker tests that expect the *base container* to be built
  * Please ensure you have a locally built version of keg base || run the command `keg base build`


mpcarolin (approved):
Looks good

daniel112 (approved):
all test passed

daniel112 (approved):
looks good to me
all test passes

PR #52, opened 2020-10-19
Keg Proxy - allows running multiple services at once

lancetipton:
**Ticket**: [ZEN-415](https://jira.simpleviewtools.com/browse/ZEN-415)
**[Semver-Status] major** 

## Context
* To allow running more then one service / repo at a time
* Allow running the taps in https
* Provide a consistent platform for running apps within a docker container
  * The proxy allows connecting to the docker container
  * Keep the containers secure
    * Most containers will have no publicly exposed ports

## Notes
* `DOC_APP_PORT` has been changed to `KEG_PROXY_PORT`
  * This is to better reflect that it's the port `keg-proxy` will route traffic to
* The service name in an apps docker-compose.yml file **MUST** be the same as the docker image name
  * This has always been the case, but I figured it would be important to mention again
* With these changes, you will **NOT** be able to load an app in the browser without the `keg-proxy` running
  * It is possible to work around this if for some reason you need to do that, but it's **NOT** setup by default
* `KEG_NO_INJECTED_COMPOSE` - Setting this env to any value will disable the injected docker-compose.yml file
  * It's currently being used for the `keg-proxy`
  * Currently there's no reason for other apps to use it
  * **99.9%** of the time setting this will break your app
  * You should **ONLY** set this if you know what you're doing and want to bypass the `keg-poxy`

## IMPORTANT
*This branch changes the general networking setup of the docker-contianers*

### The Old (Lame) Way
* Any time a container was started it would bind `DOC_APP_PORT` to `port 80` of the host machine
* If you tried to start another container after `port 80` was already bound it would throw an error
* You could only run one `container / service` at a time

### The New (Cooler) Way
* Any time a container is started, it checks if the `keg-proxy` container is running
  * If it's not, then it starts it first before starting the original container
  * Only the `keg-proxy` container is bound to `port 80` of the host machine
  * By default no other containers expose ports to the host machine
    * This can be changed using standard port mapping in the `docker-compose.yml` file
* When a container starts up it auto-registers with the `keg-proxy` container
* Labels are dynamically created on the docker container
  * These labels tell the proxy how to properly route traffic to the container
* Included in these labels is the port and host name for the container
  * The port label tells the proxy what port to send traffic to
    * This can be defined by setting the `KEG_PROXY_PORT` env in the values.yml file
    * **THIS ENV IS REQUIRED** for the container to receive any traffic
    * No default is setup because it is almost always different for each application / container
  * The host label tells the proxy what `host header` should route traffic to the container
    * This can be defined by setting the `KEG_PROXY_HOST` env in the `values.yml` file
* The general flow works like this
  1. Run the command to start the container
    1. `keg-proxy` container is automatically started if it's not running
    2. Once `keg-proxy` is started, it listens for new containers
  2. The proxy labels are automatically generated for the container
    1. Uses the `KEG_PROXY_PORT` to create the port label
    2. Uses `KEG_PROXY_HOST` to create the host header label
  3. Once the container starts, `keg-proxy` see this, and uses the labels to register the container
  4. The container is now accessible from a web browser

### App URLs and the KEG_PROXY_HOST env
* The ENV `KEG_PROXY_HOST` is not required to be defined like the `KEG_PROXY_PORT` env
  * Most of the time we already know what the host url should be
  * It follows the general pattern of
    * `<container-context-name>-<git-branch-name>.${KEG_PROXY_HOST}`
  * **Pattern Breakdown**
    * `<container-context-name>` - Context name of the container relative to the `keg-cli`
      * This name can come from a few different places
        * **Internal Apps** - Use the `keg-cli` context
          * For Example:
            * `keg-core` === `core`
              * Assuming the branch is `develop`, the full url would be `core-develop.local.kegdev.xyz`
            * `keg-components` === `components`
              * Assuming the branch is `develop`, the full url would be `components-develop.local.kegdev.xyz`
        * **Taps** - Use the linked tap name in the `keg-cli` global config
          * For Example: `events-force` tap, linked as `evf` to the `keg-cli`, the context name would be `evf`
            * Assuming the branch is `develop`, the full url would be `evf-develop.local.kegdev.xyz`
          * **IMPORTANT**
            * **Non-Injected** container name will still be `tap`, but `tap.local.kegdev.xyz` will **NO** longer resolve
    * `<git-branch-name>` - The current git banach of the container being run
      * This means, the repository is **REQUIRED** to be a `git repository` 
    * `KEG_PROXY_HOST` - Url for accessing docker containers
      * Defaults to `local.kegdev.xyz`
      * Can be overridden from the `values.yml` file
      * Generally speaking you should **NOT** need to ever set this env
  * **Notes**
    * Docker package urls follow the same pattern
    * They use the labels from the docker image at the time that they were created
    * For Example
      * `keg-core`, With a docker package like this => `docker.pkg.github.com/simpleviewinc/keg-packages/keg-core:feature-branch`
      * The url would be `core-feature-branch.local.kegdev.xyz`
  * **IMPORTANT**
    * Due to the above changes, the following URLs will no longer resolve
      * `core.local.kegdev.xyz`
        * Instead use `core-<current-local-branch>.local.kegdev.xyz`
      * `components.local.kegdev.xyz`
        * Instead use `components-<current-local-branch>.local.kegdev.xyz`
      * `tap.local.kegdev.xyz`
        * Instead use `<linked-tap-name>-<current-local-branch>.local.kegdev.xyz`

## Goal
* Allow running more then one container at a time
* Eventually run the taps in https
  * Still need to setup SSL certificate

## Updates
* `repos/keg-cli/containers/*`
  * Removed Caddy envs
  * Added Keg-Proxy envs
    * Also added documentation within the values files for the `ENVs`
  * Removed exposed ports
    * All traffic should be routed through the proxy
    * You can still expose ports, but generally they should not be
      * Exposing ports makes the containers insecure
  * **DOC_APP_PORT** has been removed, and replaced with **KEG_PROXY_PORT**
    * This better reflects the purpose of this ENV
    * This is **REQUIRED** to allow `keg-proxy` to route traffic to the container
      * The container will still run, but can not be accessed from a web browser
  * Remove volume mounts that were causing issues in some cases
* `repos/keg-cli/containers/proxy/docker-compose.yml`
  * Added for building and running the keg-poxy container
  * Still needs some work when running in a production ENV
    * There's currently no need for us to do this, so the work can be pushed to another time
* `repos/keg-cli/scripts/setup/defaults.env`
  * Added more defaults for the `keg-proxy`
* `repos/keg-cli/src/__mocks__/*`
  * Added a number of mocks to support tests for the `keg-cli`
* `repos/keg-cli/src/constants/constants.js`
  * Added mapping for `keg-proxy`
* `repos/keg-cli/src/constants/docker/index.js`
  * Added `KEG_DOCKER_NETWORK` constant, which gets applied to containers
    * This allows the `keg-proxy` access to the container for routing
* `repos/keg-cli/src/constants/docker/labels.js`
  * Added constants for docker labels to be added to all containers built with the `keg-cli`
  * Added constatns for required [traefik](https://traefik.io/traefik/), which is what the keg-proxy uses behind the scenes
* `repos/keg-cli/src/constants/docker/values.js`
  * Updated how internal container folders are found within the `keg-cli`
  * This is required because the `keg-proxy` does not have a `Dockerfile`
* `repos/keg-cli/src/constants/envs.js`
  * Updated the `keg-cli defaults.env` file to be loaded along with the `Global defaults.env`
  * This allows ensuring all ENVs get loaded even if they are not defined in the `Global defaults.env`
  * `Global defaults.env` will override values in the `keg-cli defaults.env`
* `repos/keg-cli/src/libs/docker/commands.js` && `repos/keg-cli/src/tasks/docker/build.js` && `repos/keg-cli/src/tasks/docker/compose/build.js`
  * Minor clean up
* `repos/keg-cli/src/libs/docker/image.js`
  * Cleaned up how a docker image is found
  * There were some cases where the the image could not be found, even when the image exists
  * This should fix those issues
* `repos/keg-cli/src/libs/docker/container.js`
  * Added inspect method for containers
* `repos/keg-cli/src/libs/docker/machine.js`
  * Removed unused items
* `repos/keg-cli/src/tasks/docker/build.js`
  * Refactored how the dockerBuildCmd is called
  * Requires extra params for building the keg-cli labels
* `repos/keg-cli/src/tasks/docker/compose/down.js`
  * Clean up task to allow properly shutting down a container / service
  * Added code to pull in the `proxyDomain`
* `repos/keg-cli/src/tasks/docker/image/run.js` && `repos/keg-cli/src/tasks/docker/package/run.js`
  * Updated to fix some issues with running individual containers
  * Added call to `addProxyOptions` to allow auto connecting the container to the proxy
  * This allows running an individual docker container, and still expose it through the proxy
* `repos/keg-cli/src/tasks/proxy`
  * Added tasks for managing the `keg-proxy`
* `repos/keg-cli/src/tasks/mutagen/create.js`
  * Minor update to fix logging when a sync is created
* `repos/keg-cli/src/tasks/tap/destroy.js`
  * Added remove option, to remove docker collateral 
* `repos/keg-cli/src/templates/injected-compose.template.yml`
  * Updated the injected-compose template to include generated labels
  * Added a number of other items to make setting up injected compose files easier
* `repos/keg-cli/src/*/__tests__/*`
  * Added more tests to a number of files
* `repos/keg-cli/src/utils/builders/buildContextEnvs.js`
  * Updated how the service name is found
  * Prevents some bugs from throwing in some cases
* `repos/keg-cli/src/utils/docker/buildDockerCmd.js`
  * Cleaned up how the docker command is built
  * Removed the port binding
  * Added label generation for the proxy/keg labels
* `repos/keg-cli/src/utils/docker/buildExecParams.js`
  * Fixed a bug
* `repos/keg-cli/src/utils/docker/compose/addProxyOptions.js`
  * Builds the options for a docker command, when running an image directly
* `repos/keg-cli/src/utils/docker/compose/buildComposeCmd.js`
  * Cleaned up the data object being passed to the injected template
  * Makes it easier to reference within the template 
* `repos/keg-cli/src/utils/docker/compose/generateLabels.js`
  * Generates the labels for docker-compose commands
* `repos/keg-cli/src/utils/docker/compose/getService*`
  * Removed a number of unused helper methods
* `repos/keg-cli/src/utils/docker/compose/loadComposeConfig.js`
  * Added param to skip throwing an error when compose file is not found
* `repos/keg-cli/src/utils/docker/getDockerArgs.js`
  * Removed port mapping
* `repos/keg-cli/src/utils/getters/getContext.js`
  * Updated to help paring images with tags
  * Refactored how context is found
* `repos/keg-cli/src/utils/getters/getLocationContext.js`
  * Refactored functions, no logic changes
* `repos/keg-cli/src/utils/helpers/mapEnv.js`
  * Minor bug fix in some cases
  * Added check for a `skipDockerExec` argument
  * If it exist, it skips running the default `exec` task on a started container
  * This is used in the `keg-proxy` to not auto connect to it once started
* `repos/keg-cli/src/utils/services/destroyService.js`
  * Fixed bug with the remove option passed in from the terminal
* Added call to make sure the proxy service is running for these services
  * `repos/keg-cli/src/utils/services/packageService.js` 
  * `repos/keg-cli/src/utils/services/restartService.js`
  * `repos/keg-cli/src/utils/services/runService.js`
  * `repos/keg-cli/src/utils/services/startService.js`
* `repos/keg-cli/src/utils/services/proxyService.js`
  * Service to make sure the proxy is running

## Testing
**Setup**
* Check out this PR locally
  * `keg && keg pr 52`
* Kill all running docker containers
  * `keg dc rm`
    * Enter the index of a container, and hit enter
    * Do this for all running containers 
  * Run `keg d clean -f` - To clean up images and containers
* Ensure no docker images or containers are exist
  * Use `keg di` && `keg dc` to see a list for each.
      * Nothing should exit

**Test 1**
* 
* Run the `events-force`
  * Make sure your are on the `develop` branch
    * `keg evf && keg br develop && keg pull`
  * Start the tap
    * `keg evf start`
  * This should do a number of things:
    * Automatically build the `base image`, becuase we removed it above in setup
    * Automatically run the `keg-proxy`
      * There is no build step for `keg-proxy`
    * Build and run the `events-force tap`
  * Once complete, In a separate terminal run the command `keg dc`
  * You should see two containers, the tap container and the proxy container
    * [Looks like this](http://snpy.in/B90s5O)
    * **The proxy container should start automatically**
      * There are tasks to start and stop the proxy but you should never really need them
  * Next run the command `keg di`
    * You should see images for `keg-base`, `keg-proxy ( traefik )`  , and the `tap`
    * [Looks like this](http://snpy.in/QDNM7M)

**Test 2**
* Navigate to [http://local.kegdev.xyz](http://local.kegdev.xyz)
  * This should load the `traefik` Gui
  * Click on the `http` button and scroll down to the services
    * You should see a list of services registered to the proxy
    * One of them should be the `events force tap`
    * [Looks like this](http://snpy.in/Msajfv)
  * In a new browser tab,
      * Navigate to [http://evf-develop.local.kegdev.xyz](http://evf-develop.local.kegdev.xyz)
      * Ensure the tap loads as expected 

**Test 3**
* In a different terminal window, start Keg-Components
  * Run `keg components start`
* It should auto-build the container, and start that app
* In the browser, navigate to `local.kegdev.xyz`
  * Click on the `http` tab again
    * You should now see a new entry for `keg-components`
    * [Looks like this](http://snpy.in/s0Koz1)
  * When a container starts, it will auto-register with the `keg-proxy`, and creates a new route
  * [It looks like this](http://snpy.in/0ytbos)
* Open another tab and navigate to [http://components-keg-proxy-trafik.local.kegdev.xyz](http://components-keg-proxy-trafik.local.kegdev.xyz)
  * Ensure keg-components app loads as expected

**NOTES**
* You may see warning when starting || stopping containers
  * [They look like this](http://snpy.in/Fpr0WQ)
  * You can safely ignore them
  * They are related to `docker-compose`, and how it manages services with the same project name
    * Typically service are created together, which is why the warning displays

**Test 4**
* At this point you should have
  * Events Force Tap running at [http://evf-develop.local.kegdev.xyz](http://evf-develop.local.kegdev.xyz)
  * Keg-Components running at [http://components-keg-proxy-trafik.local.kegdev.xyz](http://components-keg-proxy-trafik.local.kegdev.xyz)
  * Keg-Proxy running at [http://local.kegdev.xyz](http://local.kegdev.xyz)
* Go to an open terminal window
* Run `keg evf sync components:install:start`
  * This should
    * Create a sync between `keg-components` and Events force container
    * Run `yarn dev` in detached mode
    * [Looks like this](http://snpy.in/LWxuNT)
* Run `keg sy`
  * You should see at least 3 mutagen syncs
    * the events force tap container
    * keg-components container
    * keg-components into the events force tap container
* Run `keg evf att`
  * This should attach to the events force container
* Run `ps aux`
  * Ensure you see the keg-components [rollup process running in list](http://snpy.in/ocNhM0)

**Test 5**
* Open the events force tap in one browser tab and keg-components in different tab
* Make a change to the code your local keg-components repo
  * This should cause both the event force tap, and the keg-components apps to auto-rebuild
    * This effectively lets you work within keg-components and a tap at the same time

**NOTES**
  * For keg-components be sure to check your terminal for the re-build
  * Depending on the change, if may not re-load the browser window
  * So you can use the terminal to ensure it rebuilt properly

**Test 6**
* With the Events Force tap running, open a new terminal window
* Run the command `keg evf pack`
  * This should create a new docker image, and push it to github
    * Ensure this command works as expected
* After it has finished pushing to github, kill the running events force container
  * `keg evf kill`
* Once the tap is killed, run the command `keg evf pack run <docker-package-url>`
  * This should run the command to pull the latest container from github 
    * You already have it locally, so it won't actually pull anything
  * Then it should start the container
  * Ensure the container runs as expected
* In the browser, navigate to `local.kegdev.xyz`
  * Click on the `http` tab again
* You should see a new entry for the running app
  * [Looks like this](http://snpy.in/g2ir0H)
  * The url should follow this pattern => `<tap-context>-<branch>.local.kegdev.xyz`

**Test 7**
* For **any** linked tap / injected app on your machine
  * For injected Apps, You will have to update the values file
    * Replace the `DOC_APP_PORT` with `KEG_PROXY_PORT`
    * Add a `KEG_PROXY_HOST` if you want to define the host name
      * It should work with or with out it
  * Run `keg <name of tap> start`
  * Once its started, navigate to `local.kegdev.xyz` in the browser
    * Click on the `http`
    * Find the url in the list and natigate to it in the browser
    * It should follow a consistent pattern => ``<tap-context>-<branch>.local.kegdev.xyz``

**Test 8**
* Run the command `keg dpg run docker.pkg.github.com/simpleviewinc/keg-packages/keg-herkin:master`
  * It should pull down the docker image, and run it
  * Ensure it runs as expected
  * Ensure it registers with the `keg-proxy`
  * Ensure it creates the proper routing and url => https://herkin-master.local.kegdev.xyz
  * Ensure you can navigate to it in the browser

**Test 9**
* In a new terminal window, run `keg cli && yarn install &&  keg cli test`
  * Ensure all tests pass


repos/keg-cli/src/utils/log/logVirtualUrl.js
```diff
@@ -5,6 +5,8 @@ const { TAP_URL } = require('KegConst/constants')
  * Helper that Gets and logs the Virtual Machines IP address
  */
 const logVirtualUrl = (context) => {
+  if(context === 'proxy') return Logger.empty()
+
   const useUrl = context
```
daniel112, line 10: we should also update this url now so it's displaying the correct url
image

daniel112 (commented):
all tests including unit tests looks good
I tested out linked taps with keg-consumer following `step 7` and it works great
i have 1 minor comment on the logging

repos/keg-cli/src/utils/log/logVirtualUrl.js
lancetipton: Ahh good point. Ill update that

lancetipton: Ok, this has been updated. Also added a few more tests for the logVirtualUrl helper
  * Had to update the buildComposeCommand a bit, to get access to the generated url
  * Also added it to the `docker package run` task, which is helpful; cause who knows what branch the image label has

daniel112: noice

daniel112 (approved):
looks great to me #

mpcarolin (commented):
Looks good and passed all testing. I have some questions and suggestions in the code, but nothing blocking.

mpcarolin (approved):

PR #54, opened 2020-10-20
Parsing cssRules safely

mpcarolin:
## Context

* The consumer of a tap may have external stylesheets with security policies restricting access to the css rules
* these can throw errors when our style parser attempts to access the `stylesheet.cssRules` property, crashing the app

## Goal

* Wrap that code in a `try-catch`, logging some useful warnings if the security error is thrown, and preventing the app from crashing

## Updates

* `repos/re-theme/src/styleParser/getCSSRules.js`
  * wrote the helper for accessing the rules safely
  * also wrote tests in the associated test file
* `repos/re-theme/src/styleParser/styleSheetParser.js`
  * updated the `styleSheetParser` to use `getCSSRules`

## Testing

* `keg consumer pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-test-consumer:master`
  * this will load `keg-test-consumer` with a version of the tap that has these changes.
  * the version of `keg-test-consumer` will also have an external stylesheet I added in that usually would throw a CORS error when the `styleSheetParser` runs
* verify that the app does not crash on init
* verify that it looks as expected
* verify it logs a warning like this: 

```
Cannot parse cross-origin restricted stylesheet from source: https://s.yimg.com/kw/engadget/mod/css/app.97d3be22f549f3ee735498c6ad908c19.css 
This stylesheet will be safely ignored, but if you need access to it, then you should either ensure the domains match or adjust the stylesheet server's security policy. keg-sessions.esm.js:1
```

* `keg pr 54 && keg retheme`
* `yarn test`
* verify all tests pass

### DIFF
--- a/repos/re-theme/src/styleParser/__tests__/getCSSRules.js
+++ b/repos/re-theme/src/styleParser/__tests__/getCSSRules.js
@@ -1,0 +1,56 @@
+import { getCSSRules } from '../getCSSRules'
+
+const mockValidSheet = {
+  cssRules: [ 1, 2, 3 ]
+}
+
+const mockRestrictedSheet = {
+  get cssRules () {
+    throw new DOMException('DOMException Message', 'SecurityError')
+  }
+}
+
+const mockExceptionSheet = {
+  get cssRules () {
+    throw new Error('GenericError')
+  }
+}
+
+describe('getCSSRules', () => {
+  const orig = console.warn
+
+  beforeEach(() => {
+    console.warn = jest.fn()
+  })
+
+  afterAll(() => {
+    console.warn = orig
+  })
+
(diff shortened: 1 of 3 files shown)


lancetipton (approved):
Looks good. Just one thing,
* We should only log the warning in dev. I hate when libraries pollute my inspector
* This is not blocking, but should probably be updated

mpcarolin:
> Looks good. Just one thing,
> 
>     * We should only log the warning in dev. I hate when libraries pollute my inspector
> 
>     * This is not blocking, but should probably be updated

Resolved in slack -- keeping for now since most taps would not use the `stylesParser`, and these warnings might be important for the EVF team to see.

PR #53, opened 2020-10-19
WIP: Update repo versions, and push to NPM

lancetipton:
## Context

* To automate updating the keg-hub repos

## Goal

* Automate updating the keg-hub repos

## Updates

* **TODO**

## Testing

* **TODO**

PR #57, opened 2020-10-22
Keg Dependency Mismatch

lancetipton:
## Context

* We need to keep the dependencies consistent across the repos in keg-hub

## Goal

* Added task to compare package.json dependencies across repos
  * This just prints out a nice list of dependency version mismatch for all keg-hub repos
  * There's [another branch that has the code to update and save the dependencies](https://github.com/simpleviewinc/keg-hub/pull/53)

## Updates
* `repos/keg-cli/package.json`
  * Added semver package
* `repos/keg-cli/src/tasks/hub/dependencies.js`
  * Added task to load all package.json and find dependency mismatches
* `repos/keg-cli/src/tasks/hub/hub.js`
  * Added dependencies task

## Testing
* Pull this branch => `keg pr 57`
* Install the dependency updates for keg-cli
  * `keg cli && yarn install`
* Run the command => `keg hub dep`
  * Ensure it prints out a list of mismatched dependencies for the keg-hub repos

### DIFF
--- a/repos/keg-cli/package.json
+++ b/repos/keg-cli/package.json
@@ -42,10 +42,11 @@
     "js-yaml": "3.14.0",
     "log-update": "4.0.0",
     "module-alias": "2.2.2",
-    "rimraf": "2.6.3",
-    "write-yaml-file": "4.0.0",
     "ngrok": "3.2.7",
-    "qrcode-terminal": "0.12.0"
+    "qrcode-terminal": "0.12.0",
+    "rimraf": "2.6.3",
+    "semver": "^7.3.2",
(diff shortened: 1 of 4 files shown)


daniel112 (approved):
works for me.
I have 1 comment on this, very minor and not blocking.

if I have a version that is not an exact version, but instead use the approximate `~` or  compatible `^` i feel like that should  trigger a Mismatch, even if the version is technically the same.
i.e `1.0.0` vs `^1.0.0` should  be a mismatch

but if we follow our rules to keep all our dependencies an exact version, this scenario should never occur. so no big deal if we don't want to include the logic for it

PR #61, opened 2020-10-28
Zen 431 keg prefix filter

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-431)

`[semver]: patch`

## Context

* Currently the useStyleTag hook uses all classes of a component to create a css selector
* This works exactly as it should, but creates very specific selectors
* This makes it hard to override the styles from outside re-theme
* Update the getSelector used in the useStyleTag hook to only use classes with the keg prefix
    * All non-keg prefixed classes should be filtered out

## Goal

* update retheme getSelector to be able to filter out by prefix tag if passed in
* also minor style update to `textToggle` its showing 2 underlines in some FireFox browsers

## Updates

* `repos/keg-components/src/theme/components/textToggle.js`
   * the underline style is supposed to only be on the text component, not the container component

* `repos/re-theme/src/styleInjector/__tests__/injectHelpers.js`
   * expanded the `getSelector` tests

* `repos/re-theme/src/styleInjector/injectHelpers.js`
   * added a third optional param to filter by prefix

## Testing

* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:develop`
* verify that all the elements only have `keg-` classnames
* there should be no `ef` or `evf` classnames

### Unit Test
* pull the branch down locally
* run `keg retheme && yarn test`
* ensure all test passes

### DIFF
--- a/repos/re-theme/src/styleInjector/injectHelpers.js
+++ b/repos/re-theme/src/styleInjector/injectHelpers.js
@@ -132,7 +132,7 @@ export const getSelector = (className, cssString, filterPrefix) => {
 
   // filter by prefix if passed in
   const filterWithPrefix = cls => {
-    return filterPrefix
+    return cls && filterPrefix
       ? cls.startsWith(filterPrefix)
       : cls
   }
--- a/repos/re-theme/src/styleInjector/injectHelpers.js
+++ b/repos/re-theme/src/styleInjector/injectHelpers.js
@@ -122,17 +122,27 @@ export const filterRules = (style, filter) => {
 /**
  * Creates a unique selector based on the passed in className and cssString
  * @function
- * @param {string} className - Original className used as a css selector
+ * @param {string|Array<string>} className - Original className(s) used as a css selector
  * @param {string} cssString - Css rules for the className in string format
- *
+ * @param {string=} filterPrefix - optional prefix to filter by
+ * 
  * @returns {string} - Hashed version of the string
  */
-export const getSelector = (className, cssString) => {
+export const getSelector = (className, cssString, filterPrefix) => {
+
+  // filter by prefix if passed in
+  const filterWithPrefix = cls => {
+    return filterPrefix
+      ? cls.startsWith(filterPrefix)
+      : cls
+  }
+
   const selector = !exists(className)
     ? false
     : isArr(className)
-      ? className.filter(cls => cls).join('.').trim()
-      : isStr(className) && className.split(' ').join('.').trim()
+      ? className.filter(filterWithPrefix).join('.').trim()
+      : isStr(className) && className.split(' ').filter(filterWithPrefix).join('.').trim()
+
 
   return selector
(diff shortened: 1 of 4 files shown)


repos/re-theme/src/styleInjector/injectHelpers.js
```diff
@@ -122,17 +122,27 @@ export const filterRules = (style, filter) => {
...
+ * 
  * @returns {string} - Hashed version of the string
  */
-export const getSelector = (className, cssString) => {
+export const getSelector = (className, cssString, filterPrefix) => {
+
+  // filter by prefix if passed in
+  const filterWithPrefix = cls => {
+    return filterPrefix
+      ? cls.startsWith(filterPrefix)
```
lancetipton, line 136: Originally the check was only doing this `className.filter(cls => cls)`
  * Which filters out any `falsy` cls values,
  * Which sometimes exist based on how they are passed in from the components

So you should also check here if the class exists, before calling the `startsWith` method
  * Because if the cls value is false, then it will throw an error due to `startsWith` not being a method of `false`

Simple fix to look like this =>
```
    return cls && filterPrefix
      ? cls.startsWith(filterPrefix)
      : cls
```

repos/re-theme/src/styleInjector/useStyleTag.js
```diff
@@ -93,7 +93,7 @@ export const useStyleTag = (style, className='', config) => {
     const { blocks, filtered } = convertToCss(style, config)
 
     // Create a unique selector based on the className and built blocks
-    const selector = getSelector(className, blocks.join(''))
+    const selector = getSelector(className, blocks.join(''), 'keg')
```
lancetipton, line 96: It would be better if this could be defined outside the context of Re-Theme
* We could put it in the theme.RTMeta object, and you could call the `useTheme` hook here to get access to it.
* Or maybe we expose a method similar to the setDefaultTheme, that sets a default prefix
* That way it could be customizable

It's no blocking, and works for now, but can you create a ticket in the Jira backlog to a add that.

lancetipton (changes requested):
Looks good. Added a couple comments, and one small thing to fix

repos/re-theme/src/styleInjector/injectHelpers.js
daniel112: gotcha. updated

repos/re-theme/src/styleInjector/useStyleTag.js
daniel112: will do.
ticket here: https://jira.simpleviewtools.com/browse/ZEN-433

lancetipton: Sweet thanks

lancetipton (approved):
Looks good to me. Thanks

PR #63, opened 2020-10-30
Zen 423 updates

mpcarolin:
**Ticket**: [ZEN-423](https://jira.simpleviewtools.com/browse/ZEN-423)

## Goal

* Add JSUtil helper functions used in this PR: https://github.com/simpleviewinc/tap-events-force/pull/81
* also add unit tests

## Updates (November 3)
* removed `areShallowEqual`, since it was duplicating the existing `shallowEqual` function
* added one more utility: `repos/jsutils/src/url/getURLParam.js`
  * this is a helper I'm using to get the value of a specific url parameter easily.
  * it wraps `queryToObj`, and safely accesses the url even if you are not on a browser that defines `document`

## Updates
* `repos/jsutils/src/array/areFrequencyEqual.js`
  * compares two arrays and checks that they have the same elements with the same frequencies, ignoring order
* `repos/jsutils/src/array/areSetEqual.js`
  * compares two arrays and checks that they have the same elements, ignoring order and frequency
* `repos/jsutils/src/array/areShallowEqual.js`
  * compares two arrays and checks that they have the same elements and order
* I attempted to make these very efficient, and both `areFrequencyEqual` and `areSetEqual` prioritize speed over memory use.

## Testing
* `areSetEqual` is used extensively in the tap PR: https://github.com/simpleviewinc/tap-events-force/pull/81
* `yarn test`
  * verify all tests pass


repos/jsutils/src/array/areSetEqual.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Array<*>} arr 
+ * @param {Array<*>} otherArr 
+ */
+export const areSetEqual = (arr, otherArr) => {
+  const [ valid ] = validate({ arr, otherArr }, { $default: isArr })
+  if (!valid) return null
+
+  if (arr === otherArr) return true
+
+  const longest = arr.length > otherArr.length
```
lancetipton, line 20: Instead of doing this compare twice, you can just do this =>
```js
  const [longest, shortest] = arr.length > otherArr.length
    ? [arr, otherArr]
    : [otherArr, arr]
```

repos/jsutils/src/array/areShallowEqual.js
```diff
@@ -0,0 +1,27 @@
...
+ * @return {boolean} true if arrays are equal in order and elements. False otherwise, null if input are not both arrays
+ */
+export const areShallowEqual = (arr, otherArr) => {
+  const [ valid ] = validate({ arr, otherArr }, { $default: isArr }) 
+  if (!valid) return null
+
+  if (arr === otherArr) return true
+  if (arr.length !== otherArr.length) return false
+
+  for (let i = 0; i < arr.length; i++) {
```
lancetipton, line 21: Instead of re-writing this logic, you can just import `shallowEqual` from `../collection/shallowEqual`

But really, this method seems kind of point less. It's just implementing the `shallowEqual` logic in another place.

Why don't you just import `shallowEqual` ?

lancetipton (changes requested):
Everything looks good, but I left a few comments on some minor changes. Thanks

repos/jsutils/src/array/areShallowEqual.js
mpcarolin: good point. I forgot we had that function! I will remove `areShallowEqual`.

repos/jsutils/src/array/areSetEqual.js
mpcarolin: updated

lancetipton (approved):
Looks good. All tests pass

PR #64, opened 2020-11-04
Add Fast-Refresh for Web

lancetipton:
**[Semver-Status] minor** 

> Important
> This PR is based on of the [keg-proxy-trafik PR](https://github.com/simpleviewinc/keg-hub/pull/52)
> So this can't be merged in until the PR is merged in
> So if you have not reviewed that PR yet, [go review that one first](https://github.com/simpleviewinc/keg-hub/pull/52)

## Context

* Added webpack [fast-refresh](https://reactnative.dev/docs/fast-refresh)
* Allows hot-swapping component files without a page reload
* Keeps state after code changes in some cases

## Goal

* Improve developer experience 
* Keep state between code changes
* Remove the need for constant browser refreshing

## Updates

* `repos/keg-core/babel.config.js`
  * Updated to cache the babel config based on env
  * Added react refresh plugin
  * Added react refresh plugin to dependencies

## Testing

**Test 1**
* This test is just to show docker packages still work with the update
  * **It does not test the added functionality**
* Run the docker package url => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:add-fast-refresh`
  * Ensure the app works as expected 

**Test 2**
* Check out the develop branch for the events-force tap
  * `keg evf && keg br develop`
* Then pull this PR => `keg && keg pr 64`
* Next, start the events force tap, with a sync for core & components
  * `keg evf start --sync core:install,components:install:start`
    * This command will start the events force tap, sync, install dependencies for core, and sync, install, run rollup for components
* Once this command completes, navigate your browser to [evf-develop.local.kegdev.xyz](http://evf-develop.local.kegdev.xyz)
  * Ensure the app works as expected in the browser
 * Now open the code for events force tap, and make a change to one of the components and save the change
  * Ensure it's a component that currently being used on the page _( evf-button is a good one )_
* In the browser you should see the change without the tab refreshing
  * In some cases the browser **WILL REFRESH**
    * It really depends on the change that was made, and if it can be hot-loaded into the app
    * If the page does refresh, try making a small change like `console.log` in a functional component
    * This typically works with hot-mounting
* Next make a change in the keg-components
  * The same rules above, apply here.
  * The change should be reflected on the page without a tab refresh
  * There are cases where the page still does a full refresh, but they should be minimal
 
**Test 3**
* In another terminal window add a sync for retheme
  * `keg evf sync retheme:install:run`
* Make a change within the retheme repo, and save it
  * The same rules above, apply here.
  * The change should be reflected on the page without a tab refresh
  * There are cases where the page still does a full refresh, but they should be minimal


mpcarolin (changes requested):
When this works, it's awesome.

However, I notice a pretty consistent error gets thrown whenever I change the _order_ or _number_ of hooks in a component.

I’ve seen this happen when I add or remove a hook (every time), or if I rearrange hooks (sometimes).

I know fast-refresh is trying to preserve the state between reloads, but the plugin here doesn’t seem to be doing it very well yet (it’s still marked “experimental”). What should be happening is it should know to do a full-refresh at that point.

I think we should wait to add this until the plugin has matured.

image

lancetipton:
Closing for now. May re-open when the plugin becomes more mature

PR #66, opened 2020-11-04
update keg prefix to still apply the classname to component

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-447)
`[semver]: patch`
> - related to this tap PR: https://github.com/simpleviewinc/tap-events-force/pull/86

## Context

* we removed non keg classnames from both the selector and classList on https://jira.simpleviewtools.com/browse/ZEN-431
* what we shouldve done is remove it form the selector only

## Goal

- non `keg` classname should only be filtered out on the selectors. NOT classList
## Updates

- injectHelpers unit tests update
- updated getSelector to return an object instead of a string
- updated useStyleTag so it would filter out the non keg selectors but still keep it in the classList

## Testing
* run `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-447-demo`
* navigate to http://evf-ZEN-447-demo.local.kegdev.xyz
* inspect the `select` button or any other components that you know have an `ef-` class on it
* verify that the `ef-` className does not exists on the selector, but does exists in the classList
   * image

* pull this branch
* run `keg retheme && yarn test`
* verify all retheme unit tests passes

### DIFF
--- a/repos/re-theme/src/styleInjector/__tests__/injectHelpers.js
+++ b/repos/re-theme/src/styleInjector/__tests__/injectHelpers.js
@@ -99,24 +99,24 @@ describe('injectHelpers', () => {
     })
 
     it('should return a css className selector with a has the passed in className', () => {
-      const selector = getSelector(`test-class`, `my-test-styles`)
+      const {selector} = getSelector(`test-class`, `my-test-styles`)
       expect(selector.includes(`test-class`)).toBe(true)
     })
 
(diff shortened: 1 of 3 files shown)


mpcarolin (approved):
Looks good to me.

lancetipton (approved):
Looks good, and working for me

PR #67, opened 2020-11-05
 Zen 356 keg comp per file

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-356)

## Context

* currently, keg-components builds a single export file per module type (esm, cjs)
* this works, but when we tried building tap-events-force, rollup struggled to tree shake everything in keg-components, so it bundled in dependencies we weren't necessarily using, like AppHeader
*to ensure consumers can properly tree shake keg-components, we should update the rollup config to export one file per component
see jsutils for an example of this already working

## Goal

* update the rollup config with multiple output files
* update keg-core webpack with the new alias
* we should be able to import @keg-components with specific file now
    * ex: `import { Button } from '@keg-hub/keg-components/button'`
## Updates

* `keg-components/configs/rollup.config.js`
   * updated to export per file for components, utils, hooks
   * theme is imported with the `theme/index.js`
   * dynamically goes through each file in `/src` and uses either the `.native.js` (if native && if it exists), otherwise will use the `.js`
* `src/*`
   * many of the file changes shown is just file/folder restructure for consistency

### Web
* **keg-core**: `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-core:zen-356-keg-comp-per-file`

* **tap-evf**: `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-356-demo`

* verify that both apps run and function as you'd expect

### Per file export
* Pull this branch
* run keg-core synced with keg-components: `keg core start --sync components:install:build`
* navigate to http://core-ZEN-356-keg-comp-per-file.local.kegdev.xyz
* open file: `keg-core/base/app.js`
* update the components import to use specific @keg-components import (Appheader, Select, Option)
   * image
* verify the app still runs and you still see those components

### Native
* covered in this PR https://github.com/simpleviewinc/keg-hub/pull/68


repos/keg-components/configs/rollup.config.js
```diff
@@ -7,12 +7,142 @@ import cleanup from 'rollup-plugin-cleanup'
...
+ * @param {boolean} isNative - if we're building for native or not
+ * @param {Array<string>} items - array containing filenames and/or folder names
+ * @param {Array<string>} ignoreList - which file/folders to ignore
+ * 
+ * @returns {object}
+ */
+const createExportMappings = (dir, isNative, items=[], ignoreList=[]) => {
+
+  const obj = {}
+  items.forEach((item) => {
```
lancetipton, line 78: Nothing wrong here, but I typically do a `reduce` here
That way I don't have to define an external variable `obj`, and I can return on the same line.
Looks like this => 
```
  return items.reduce((obj, item) => {
    // ... all the codes

    return obj
  }, {})
```

repos/keg-components/configs/rollup.config.js
```diff
@@ -7,12 +7,142 @@ import cleanup from 'rollup-plugin-cleanup'
...
+
+ /**
+  * setup the input files for specific exports
+  * @param {string} platform - web or native
+  * 
+  * @returns {object} - object containing specific exports
+  */
+const getInputs = (platform) => {
+  return {
+    svgIcons: 'src/assets/icons/svgIcon',
```
lancetipton, line 140: This works, but really `svgIcon` should be moved to the components folder. 
* I know I put it where it's at currently, but it's wrong
* Not saying you need to change it with this PR, but just making a not of it

repos/keg-components/src/components/typography/caption/index.js
```diff
@@ -0,0 +1 @@
+export * from './caption'
```
lancetipton, line 1: I'm not sure I follow the need for doing this?
Not just in this file, but for this and all the rest of them?

repos/keg-components/src/hooks/useClassList/useClassList.native.js
```diff
@@ -1,4 +1,4 @@
-import { noPropArr } from '../utils/helpers/noop'
+import { noPropArr } from '@keg-hub/jsutils'
```
lancetipton, line 1: Why update this one, and not the test of them?

lancetipton (changes requested):
Overall looks good, so really great work. All the tests in the PR pass, and everything seems to work correctly, but `yarn test` is failing
* I have a fix in this [PR](https://github.com/simpleviewinc/keg-hub/pull/69) for it.
* Pretty sure it's not related to your changes
* I have a few questions:
  * Some of the folder structure changes do make sense? Why move some many files into folders
  * Some of the noop imports where updated to point to jsutils, but not all of them. Any reason for this?
  * I left a few comments on how I do things, but that does not mean you need to update it. Just figured I'd let you know.

Marking as `request changes`, but really I'm just looking for answers to the above questions.
Thanks

repos/keg-components/src/hooks/useClassList/useClassList.native.js
daniel112: I mightve just missed it cos I had so many changes. I will look to update all of them

repos/keg-components/src/components/typography/caption/index.js
daniel112: on this comment
> Some of the folder structure changes do make sense? Why move some many files into folders
I'm gonna guess you mean it does **not** make sense?

1. I think originally I was grabbing the file based on the `index.js` file so the file structure being what it is now really helps with that process (and it was somewhat consistent). 
2. Now that I'm grabbing the file a different way, the file structure wont help or hurt my process. so i can revert most of this back.
3. i think for files that has a .native counterpart. it's still helpful to put them in a folder. so I think i want to keep those, what do you think?
    - ex: `/useAnimate/useAnimate.js` && `/useAnimate/useAnimate.native.js`

repos/keg-components/src/hooks/useClassList/useClassList.native.js
daniel112: updated

repos/keg-components/src/components/typography/caption/index.js
daniel112: updated to only move file into folder if it has a .native counterpart

repos/keg-components/configs/rollup.config.js
daniel112: updated

repos/keg-components/configs/rollup.config.js
daniel112: moved it to /src/components

lancetipton (approved):
Looks good, and works great

Thanks

PR #72, opened 2020-11-12
Keg-Components - Theme as Functions

lancetipton:
[Semver-Status] major

> This PR should be reviewed along with [this PR](https://github.com/simpleviewinc/tap-events-force/pull/92)

##  Context

* Currently there is no way to override the theme defaults of keg-components
* Any dynamically generated values or states, must be replicated manually

## Goal

* Allow overwriting the theme defaults by converted all theme objects to functions
* The defaults are then pulled in when the function is called instead of on initial load
* This allows overwriting the default theme config before build the theme object

## Updates

* `repos/keg-components/.storybook/preview.js`
  * Added call to initialize theme
* `repos/keg-components/src/theme/*`
  * Converted all theme files into functions
  * Updated on theme initialization to call each theme function
  * Allows passing in a config object
  * Allow overwriting the theme defaults 
* `repos/keg-components/src/*/*.stories.js`
  * Updated the hierarchy
  * Cleaned up layout in storybook
  * Added sort helper
 

## Testing

* Run the docker package => `keg dpg run docker.pkg.github.com/simpleviewinc/keg-packages/keg-components:keg-theme-as-functions`
* Navigate to this URL => http://components-keg-theme-as-functions.local.kegdev.xyz
* Ensure all components work as expected
  * There should be no difference between the components with these changes, and how they were before


repos/keg-components/src/theme/components/button/contained.js
```diff
@@ -1,45 +1,52 @@
...
-          : get(colors, 'palette.white01'),
-        fontSize: 14,
-        fontWeight: '500',
-        letterSpacing: 0.5,
-        textAlign: 'center',
+export const containedInit = (config=noOpObj) => {
+  const { colors, states } = getThemeDefaults()
+  const __transition = transition(config)
+
+  const containedStyles = (state, colorType) => {
```
daniel112, line 10: not related to your changes here since im noticing it on develop as well, not sure when this bug was introduced, but for `button` components:
- there's a bug where if you click+drag on a button, it removes some of the styling, until you hover on it again
buttons

- for comparison, I looked at the old [keg-components gh page](https://simpleviewinc.github.io/keg-components/?path=/story/button-outline--standard) and it's not doing it there

I can spin off a ticket for this

daniel112 (approved):
components all work as expected
I really dig the intro docs

I left a few comments but none is blocking

repos/keg-components/src/components/image/image.stories.js
```diff
@@ -5,7 +5,7 @@ import { Image, Link, Text } from '../../'
 
 const wrapStyles = { textAlign: 'center' }
 
-storiesOf('Display/Image', module)
+storiesOf('Components/Display/Image', module)
```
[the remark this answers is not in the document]
daniel112: ticket made: https://jira.simpleviewtools.com/browse/ZEN-461

repos/keg-components/src/theme/components/button/contained.js
daniel112: ticket: https://jira.simpleviewtools.com/browse/ZEN-462

repos/keg-components/src/kegComponents.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+```
+
+## Import / Use 
+
+This library allows importing just what you need or the entire thing. Each component is exported individually to help keep bundle sizes down
+
+**Individually**
+
+```js
+import { Button } from '@keg-hub/keg-components/button'
```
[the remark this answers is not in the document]
lancetipton: Typically this should not be needed. Me mostly do this because of the Keg-Hub setup. 
But to be honest, it's been a while since I tested it without the that.
Might be worth doing, the key point here is, you should not need to do this for a normal app, consuming keg-components

repos/keg-components/src/theme/components/button/contained.js
lancetipton: Hmm, this is interesting. I did a little research on it.
Looks to be related to the latest update to [react-native-web 14](https://github.com/necolas/react-native-web/issues/1719)
  * There were a few changes, that were "fixed", but really didn't fix things for us
    * Potentially related Issues
      * https://github.com/necolas/react-native-web/issues/1264
      * https://github.com/necolas/react-native-web/issues/1604
      * https://github.com/necolas/react-native-web/issues/1640

daniel112: added those links to the ticket

mpcarolin (approved):
Looks good to me

PR #73, opened 2020-11-13
V7.0.0

daniel112:
repos:
- keg-core
- keg-components
- re-theme
- jsutils
-tap-resolver


lancetipton (changes requested):
Looks good, but can you do new build for keg-components as well. That way the latest is included in with the repo. Thanks

lancetipton (approved):
Thanks

PR #74, opened 2020-11-16
Fix proxy tunnel

lancetipton:
## Context

* With the changes to use the key-proxy, the proxy tunnel no longer works

## Goal

* Allow the proxy tunnel to work with the keg-proxy

## Updates

* `repos/keg-cli/src/libs/docker/helpers.js`
  * Added helper to convert string labels from docker images into a `labels` object
* `repos/keg-cli/src/tasks/docker/tunnel.js`
  * Cleaned up code to allow running a tunnel with the keg-porxy
  * Will now ask for which container to proxy when a host is not passed as an option
  * Updated to call ngrok binary directly instead of the node wrapper
    * The node wrapper didn't allow all options
    * Added basic logging, but this could be improved if needed
* `repos/keg-cli/src/utils/getters/getInternalIp.js`
  * Small fix to a test, to filter out `Link-Local` Ips

## Testing

**Setup**
* Switch to this PR => `keg && keg pr 74`
* Run the docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:keg-theme-as-functions`
* Navagate to this URL to ensure it's working => http://evf-keg-theme-as-functions.local.kegdev.xyz
  * Ensure the tap is loading

**Test 1**
* Open a new terminal tab
* Run command `keg d tunnel`
  * It should ask you to select a container
  * Select the `package-tap-keg-theme-as-functions` container
* It should print out `starting tunnel`
* It should then print out the Tunnel URL
  * [Looks like this ](http://snpy.in/Qh3EKZ)
  * Copy the Tunnel link into the browser
    * It should route to the container
  * Ensure the tunnel url works as expected 

**Test 2**
* Stop the running tunnel from test 1
* Run this command `keg d tunnel invalid-host`
  * It should do exactly the same thing as it did in **Test 1**
  * **Except** the source host should be http://invalid-host.local.kegdev.xyz:80
    * [Looks like this](http://snpy.in/2hcZ5N)
* If you try to navigate to the host is will fail to load the tap
* This test proves
  * The passed in option overrides the default
  * It will not ask you to select a container when you pass in the host option

**Test 3**
* Stop the running tunnel from test 2
* Run this command `keg d tunnel evf-keg-theme-as-functions`
  * It should do exactly the same thing as it did in **Test 1**
  * Go to the printed out tunnel url in the browser
  * Ensure it works as expected, because the passed in host is correct

**Test 4**
* Run command `keg cli test`
  * Ensure all tests pass

### DIFF
--- a/repos/keg-cli/src/libs/docker/helpers.js
+++ b/repos/keg-cli/src/libs/docker/helpers.js
@@ -171,7 +171,18 @@ const jsonOutput = (data, skipError) => {
           built.rootId = built.repository.indexOf('/') !== -1
             ? built.repository.split('/').pop()
             : built.repository
+        else if(built.image && isStr(built.labels)){
 
+          // Convert container labels from a string to an object
+          built.labelsObj = built.labels.split(',')
+            .reduce((labelObj, label) => {
+              const [ key, value ] = label.split('=')
+              key && value && (labelObj[key] = value)
+
+              return labelObj
+            }, {})
+
+        }
         const existing = indexMap[built.id] && items[ indexMap[built.id] ]
 
         // De-dupes the returned images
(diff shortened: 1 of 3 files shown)


mpcarolin (approved):
Looks good. Works great, exactly as I expected. Tested on mobile and desktop.

daniel112 (approved):
looks good to me

nnormansv:
nice work.  looks good to me!

PR #75, opened 2020-11-18
Fix Theme Import

lancetipton:
## Context

* Theme was importing from the wrong file for the keg-theme-as-functions branch
* This fixes that import

## Goal

* To fix the theme import

## Updates

* `repos/keg-core/core/base/theme/index.js`
  * Fixed theme import

## Testing
* Run the docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:develop`
* Navigate to http://evf-develop.local.kegdev.xyz/ in the browser
  * Ensure the site runs as expected

### DIFF
--- a/repos/keg-core/core/base/theme/index.js
+++ b/repos/keg-core/core/base/theme/index.js
@@ -1,1 +1,1 @@
-export * from 'SVTheme/tapIndex'
+export * from './__kegTheme'


mpcarolin (approved):
Looks good

PR #76, opened 2020-11-19
Log Branch on Image/Container Removal

mpcarolin:
## Context

* whenever I run `keg di rm` or `keg dc rm`, I can't tell which branches those images/containers are using
* I think it's helpful to know, so you don't delete an image or container you actually want to keep

## Goal

* log out the branch in these commands

## Updates
* `repos/keg-cli/src/utils/docker/containerSelect.js`
  * updated to parse the labels domain value for logging out
* `repos/keg-cli/src/utils/docker/imageSelect.js`
  * updated to print out the tag 

## Testing
* `keg && keg pr 76`
* `keg di rm`
  * verify that the branch is logged out in the tag column
* `keg dc rm`
  * verify that the domain (which includes the branch) is logged out in the domain column

### DIFF
--- a/repos/keg-cli/src/__mocks__/libs/docker/docker.js
+++ b/repos/keg-cli/src/__mocks__/libs/docker/docker.js
@@ -354,6 +354,9 @@ const docker = {
             return data.id === image || data.image === image || data.name === image
           })
     }),
+    list: jest.fn(() => {
+      return Object.values(global.testDocker.images)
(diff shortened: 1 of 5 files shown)


lancetipton (approved):
Looks good. I made a few minor updates
  * Update containerSelect to use ENVs and constants
  * Added tests for both containerSelect and imageSelect

PR #78, opened 2020-12-01
implemented parse error util

mpcarolin:
## Updates

* `repos/jsutils/src/method/parseErrorMessage.js`
  * wrote a function for safely parsing the message from an exception, regardless of its type (string or object)
  * added unit test

## Testing

* `yarn test`


lancetipton (approved):
Looks good. Thanks

PR #81, opened 2020-12-04
Jsutils more helpers

lancetipton:
## Context

* Saw some common code patterns that could be extracted to utility methods

## Goal

* Add some utility methods

## Updates

* repos/jsutils/src/method/hasDomAccess.js
  * Helper to check if in a browser / Dom context
* repos/jsutils/src/method/__tests__/hasDomAccess.js
  * Tests for utility method
* repos/jsutils/src/string/spaceJoin.js
  * Helper to join two strings or an array of strings together
* repos/jsutils/src/string/__tests__/spaceJoin.js
  * Tests for utility method

## Testing

* Pull pr => `keg pr 81`
* Run tests => `keg hub run test jsutils`


repos/jsutils/src/string/__tests__/spaceJoin.js
```diff
@@ -0,0 +1,32 @@
...
+  it('should handle non-strings gracefully', () => {
+    expect(spaceJoin(0, defStr)).toBe(defStr)
+    expect(spaceJoin([], defStr)).toBe(defStr)
+    expect(spaceJoin({}, defStr)).toBe(defStr)
+    expect(spaceJoin(true, defStr)).toBe(defStr)
+    expect(spaceJoin(false, defStr)).toBe(defStr)
+    expect(spaceJoin(undefined, defStr)).toBe(defStr)
+    expect(spaceJoin(null, defStr)).toBe(defStr)
+  })
+
```
daniel112, line 31: should we also have a test where both or second params are non strings? just to cover all the basis

daniel112 (approved):
looks good. test passes.
have a comment but its not blocking

repos/jsutils/src/string/__tests__/spaceJoin.js
lancetipton: Good catch, I'll add some checks for that

PR #84, opened 2020-12-04
fixed bug with missed keg prefix

mpcarolin:
## Context

* Checkbox items on the tap were greyed out when the `close` property was true, ultimately due to a bug in keg-components

## Goal

* fix this bug

## Updates

* `repos/keg-components/src/hooks/useThemeTypeAsClass/useThemeTypeAsClass.js`
  * ensured that it only includes the surface in the class array if it is a member of the valid surfaces

## Testing
* `keg components pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-components:class-prefix-fix`
  * ensure everything looks right in storybook with checkboxes
* `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-423-pending-sessions`
  * open a group booking modal, and verify that the checkboxes do not look greyed-out


lancetipton (approved):
Looks good, thanks

PR #90, opened 2020-12-08
CLI-git-lib-updates

lancetipton:
**Ticket**: [ZEN-463](https://jira.simpleviewtools.com/browse/ZEN-463)

[Semver-Status] patch

## Context

* These changes are part of a bigger ticket to update how docker images are managed. 
* These changes were extracted to to allow for a smaller PR. Because of this; some functionality may not be used in this PR

## Goal

* Removed un-used docker volumes constants
* Improve git lib's fetch functionality
* Fix bug with args-parse array parsing

## Updates

* `repos/args-parse/src/options/checkValueType.js`
  * Fixes a bug when option type is array, but the value is invalid, or does not exist
* `repos/keg-cli/src/constants/docker/volumes.js`
  * Removed unused file
* `repos/keg-cli/src/libs/git/*`
  * Added helper to list remote branches
  * Added helper to print the current git branch commit hash
  * Updated fetch helper to auto-track fetched branches

## Testing

**Setup**
* Pull this PR => `keg cli && keg pr 90`
* Install the latest deps => `yarn install`
* Copy the args-parse changes into the keg-cli node_modules
  * `cp ~/keg-hub/repos/args-parse/src/options/checkValueType.js ~/keg-hub/repos/keg-cli/node_modules/@keg-hub/args-parse/src/options/checkValueType.js`
  * This ensures you can test the args-parse changes

**Test 1**
* Run the command `keg fetch`
  * This should fetch all branches from the origin remote, and automatically set them to track their remote counterpart

**Test 2**
* Run the command `keg cli test`
  * Ensure all tests pass


repos/keg-cli/src/libs/git/repo.js
```diff
@@ -64,18 +64,39 @@ class Repo {
...
+    // Get the remote branches to setup tracking
+    const remoteBranches = await this.git.branch.listRemote(location)
+
+    // Loop over each remote branch and set it up to be tracked by the local branch
+    return Promise.all(
+      remoteBranches.map(branch => {
+        try {
+          return gitCli({
+            opts: `branch --track ${branch.name} ${branch.remote}/${branch.name}`,
+            ...options,
```
daniel112, line 93: this throws a silent error that `options` is not defined, so i dont think this git cmd is successful on any of the branch

this cmd is suppose to create local branches that I don't have and tie it to the remote right? it is currently not doing that.

daniel112 (changes requested):
unit test passed. just noticed 1 minor thing

repos/keg-cli/src/libs/git/repo.js
lancetipton: This is fixed

daniel112 (approved):
looks good to me

PR #103, opened 2020-12-18
ZEN-509: version display

mpcarolin:
**Ticket**: [ZEN-509](https://jira.simpleviewtools.com/browse/zen-509)

## Context / Goal

* to help inform devs (and anyone testing the consumer) of the current version of the tap, we should show the version of the tap in the consumer that also links to the release notes
* I thought this should go in the tap, so that the events-force team could enable it on their end too, for their testers.
* we also should update the core to support setting envs using the app configs (e.g. `tap.js`). Since that could help with this implementation, I'm doing it in this branch

## Updates

* `repos/keg-core/app.json`
  * updated to define some default envs
* `repos/keg-core/core/base/components/meta/versionDisplay.js`
  * implemented the version display component and exported it here so any tap can use it
  * updated to use `DefinePlugins` with the keg envs
* `repos/keg-core/core/scripts/js/getKegEnvs.js`
  * script for getting all the envs defined by tap and core
* `repos/keg-core/core/scripts/js/printKegEnvs.js` & `repos/keg-core/core/scripts/js/printKegEnvs.js`
  * script for printing out the env values. Helpful for debugging

### Setup
keg-test-consumer: `keg consumer pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/keg-test-consumer:master`
  * loads the consumer with the tap and core synced in
  * this will test that the rollup config works properly
tap-events-force: `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:version-display`
  * loads the tap with the core synced in
  * this will test that the webpack config works properly
* run both packages and verify:
  * you see the version display component at the bottom right of the window
  * it appears correctly on mobile web as well
  * the version display button shows `v2.3.1` (if it does, this proves the environment variables are working)
  * when you select the button, it should open the release notes for `v2.3.1`
* in your terminal, run `keg evf att` to attach to the tap package
* `yarn print:envs`
* verify the output looks like this:
image

* if you are inclined to test out more, pull these branches down (core: `ZEN-509-version-display`; tap: `version-display`) and add/modify/remove any envs from the core's `app.json` or the tap's `tap.js`


repos/keg-core/core/base/components/meta/versionDisplay.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const openHomepage = useOpenHomepageCallback(homepage, version)
+
+  const buttonStyles = useMemo(() => ({ 
+    ...versionStyles, 
+    ...styles 
+  }), [ styles ])
+
+  return (
+    <Button
```
lancetipton, line 52: You could have just used the `Link` component here. It has a number of benifits
  * It also uses the `Touchable` component just like the Button
  * You don't need to import the `Linking` library
  * You can just pass in the URL as an `href` prop
  * You can have it open in a new tab instead of the current one

That being said, you could also just
  * Make the content a `Text` component.
  * Then add `target="_blank"` and `href` props
  * That way you could keep the styles

repos/keg-core/core/scripts/js/getKegEnvs.js
```diff
@@ -0,0 +1,88 @@
...
+const buildEnvs = replaceContext => {
+  const tapEnvs = get(replaceContext, 'tap.config.keg.envs')
+  const coreEnvs = get(replaceContext, 'core.config.keg.envs')
+
+  const envs = {
+    ...coreEnvs,
+    ...tapEnvs
+  }
+
+  const getFallback = value =>
```
lancetipton, line 43: Not a big deal, but I typically don't create inline functions, unless I have to.
It's easier to maintain long term, and is consistent with single purpose functions methodology. Which I try to practice, but don't always succeed.

Plus if you ever need the method some where else, it's easier to make it reusaeable

repos/keg-core/core/scripts/js/getKegEnvs.js
```diff
@@ -0,0 +1,88 @@
...
+  const tapEnvs = get(replaceContext, 'tap.config.keg.envs')
+  const coreEnvs = get(replaceContext, 'core.config.keg.envs')
+
+  const envs = {
+    ...coreEnvs,
+    ...tapEnvs
+  }
+
+  const getFallback = value => 
+    isStr(value) && (value.startsWith('tap.') || value.startsWith('core.'))
```
lancetipton, line 44: I'm not sure I understand the point of this?
Why not just use `value` as the fallback
If it doesn't exist, then it will be `undefined` anyways

repos/keg-core/core/scripts/js/getKegEnvs.js
```diff
@@ -0,0 +1,88 @@
...
+      ? undefined
+      : value
+
+  return Object.entries(envs).reduce(
+    (replacements, [key, value]) => {
+      const replacement = get(replaceContext, value, getFallback(value))
+      return {
```
lancetipton, line 51: I used to do reduce loops this way, but then I stopped because it's creating a new object on every loop
It's not a huge deal, but is a minor performance improvement.
Now I just set the key / value to the object and return it

repos/keg-core/core/scripts/js/getKegEnvs.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/**
+ * @param {string} tapPath - path to the current tap's root
+ * @param {string} corePath - path to keg-core's root
+ * @return {Object} - all replacement definitions from the tap and core,
+ * merged into one object, compatible with plugins like `@rollup/plugin-replace`
+ * and `webpack.DefinePlugin`
+ * 
+ * To define a new env, edit your tap config (e.g. tap.js) with a new `keg.envs`
```
lancetipton, line 68: We should also add this to the readme some where, so people know about it

repos/keg-core/core/scripts/js/printKegEnvs.js
```diff
@@ -0,0 +1,19 @@
+const { getKegEnvs } = require('./getKegEnvs')
```
lancetipton, line 1: Ahh nice. I love little helpers like this. Good stuff

repos/keg-core/core/base/components/meta/versionDisplay.js
```diff
@@ -0,0 +1,75 @@
...
+  () => {
+    if (!homepage) return
+
+    const versionEndpoint = version ? `/v${version}` : ''
+    const uri = `${homepage}/releases/${versionEndpoint}`
+
+    Linking
+      .canOpenURL(uri)
+      .then(supported => supported 
+        ? Linking.openURL(uri)
```
lancetipton, line 21: Is there a way to make this open in a new tab?
When testing, it used the same tag :(

lancetipton (changes requested):
Everything looks good. 
Just one comment about opening in a new tab

repos/keg-core/core/scripts/js/getKegEnvs.js
mpcarolin: In the case of the dynamic values, like `tap.package.version`, if there isn't a value for it, then returning `value` is just returning `tap.package.verson`, which is not undefined. This just makes sure that it does.

lancetipton: Ahh got it. Thanks

repos/keg-core/core/base/components/meta/versionDisplay.js
mpcarolin: I actually forgot about the link component in keg-components. I'll just update it to use that.

repos/keg-core/core/scripts/js/getKegEnvs.js
mpcarolin: I'll update the variable names here to try to be clearer.

repos/keg-core/core/base/components/meta/versionDisplay.js
lancetipton: word

repos/keg-core/core/scripts/js/getKegEnvs.js
mpcarolin: I usually don't, but I think this function is so specific to its parent function that it makes most sense to keep it in this scope.

However, thinking about that some more, I realized I don't need it to be a function at all. I can just do that work inside the reduce loop, with less overall code.

repos/keg-core/core/scripts/js/getKegEnvs.js
mpcarolin: Man, I really like doing it this way. I guess because it's "immutable", and there's a new object every time. 

But you're right, it is slower, and there's nothing I hate more than a really long build step, so I'd rather not contribute to that! 

Updated

repos/keg-core/core/scripts/js/getKegEnvs.js
mpcarolin: 100% agree. Updated to add a whole section on envs.

We also should add a section for the tap config in general at some point.

repos/keg-core/core/base/components/meta/versionDisplay.js
mpcarolin: changed to a link component with `target={'_blank'}`, so this now works

repos/keg-core/core/base/components/meta/versionDisplay.js
mpcarolin: updated to use Link 👍🏼

daniel112 (approved):
looks good to me

lancetipton (approved):
Everything looks good. All test pass. Thanks

PR #138, opened 2021-02-17
Autocomplete component + ReStyle Fix

mpcarolin:
## Context

* We implemented an `Autocomplete` component ages ago for `tap-zerista`, and I kept kicking a migration down the road 
* There's no more road left so I'm migrating it now
* This should be used in the keg-herkin frontend

## Goal

* Migrate the autocomplete component over, making modifications where necessary to be compatible with our latest keg-components and utilities
* Implement story examples
* Fix a bug in retheme's `reStyle` function in which any update to the styles object would not rerender the component
* Fix a bug in our `Input` component that caused it to lose focus when used as a controlled-component

## Updates

* `repos/keg-components/src/components/form/autocomplete/*.js`
  * root autocomplete component
  * story examples
* `repos/keg-components/src/components/form/scrollable/select/scrollableSelect.js`
  * The scrolling list of possible autocomplete values
* `repos/keg-components/src/components/form/scrollable/item/item.js`
  * the individual row items in the scrollable menu list for autocomplete
* `repos/keg-components/src/components/form/scrollable/select/selectView.js`
  * just a styled view used in the scrollable select to wrap the menu items
* `repos/keg-components/src/components/internal/input.js`
  * removed the Touchable wrapper which was causing the focus-lost bug
  * as far as I can tell, we really don't need the touchable wrapper
  * we aren't using the touchable props like `onPress` prop anywhere in `tap-events-force` or `keg-components`
    * but just in case, I let the `onPress` prop work with `onFocus`
* `repos/keg-components/src/hooks/useAutocompleteItems.js`
  * the hook that does the actual autocompleting work
  * finds the subset of all possible values that overlaps with the user's input
* `repos/keg-components/src/utils/events/getTextFromChangeEvent.js`
  * cross platform helper for getting text from a change event
* `repos/re-theme/src/reStyle/reStyle.js`
  * removed the use of `useObjWithIdentity` in `reStyle`, replacing it with a new hook: `useShallowMemoMerge`
* `repos/re-theme/src/reStyle/reStyleHooks.js`
  * added `useShallowMemoMerge` and unit tests

## Testing

1. `keg dpg run keg-components:autocomplete-component`
2. Navigate to http://components-autocomplete-component.local.kegdev.xyz/?path=/story/components-form-autocomplete--basic
3. Click on the input element, and start typing "Alpine" (or anything really, but Alpine has two branches)
3a. Verify that the menu list appears and is superimposed over the text "Click on an item to select it"
4. Select one of the options in the list
5. Verify that 
    * you see the option you selected in the "Actions" storybook tab
    * what you selected is filled into the input field
6. Now, clear out what you typed, and type in "Lance"
7. Hit the enter key
8. Verify the conditions listed in step 5
9. Clear the field again, then type just "a"
10. Verify you can scroll through the list to see all the items with the letter "a"
11. Verify you can use arrow keys to highlight different items in the list, and that you can press enter to select one
12. Verify that when you click outside of the scrollable list, it hides


repos/keg-components/src/components/form/autocomplete/autocomplete.example.js
```diff
@@ -0,0 +1,44 @@
...
+  'Saanen',
+  'Toggenburg',
+  'Russian White',
+  'Valais Blackneck',
+]
+
+export const Basic = () => {
+  return (
+    <View>
+      <p>Start typing to see options</p>
```
lancetipton, line 29: Can you use our own `P` component. How can we expect others to use it, if we don't even use it?

repos/keg-components/src/components/form/scrollable/item/item.js
```diff
@@ -0,0 +1,58 @@
+import React, { useCallback } from 'react'
+import { Button } from 'KegButton'
+import { reStyle } from '@keg-hub/re-theme/reStyle'
+import { noOp } from '@keg-hub/jsutils'
+import PropTypes from 'prop-types'
+
+/**
+ * A Button styled for the autocomplete
+ */
+const SelectButton = reStyle(
```
lancetipton, line 10: Seems like we should allow the props.styles to override these, unless I'm missing something?
Simple way to do that, is just do a `deepMerge` with `props.styles` and the final object on line 29.
We should  still be able to override them by passing it to the final component, but I have not verified that. If that's the case, then this is fine.

repos/keg-components/src/components/form/scrollable/select/scrollableSelect.stories.mdx
```diff
@@ -0,0 +1,31 @@
+<!--- scrollableSelect.stories.mdx -->
```
lancetipton, line 1: I'm getting an error when I try to view this in StoryBook => http://snpy.in/XMACUc

repos/keg-components/src/components/form/scrollable/select/scrollableSelect.example.js
```diff
@@ -0,0 +1,54 @@
...
+    visible={true}
+    items={colors.map(clr => ({ text: clr }))}
+    onSelect={({ text }) => console.log('selected ', text)}
+  />
+)
+
+export const AdjustableHeight = () => {
+  const [ height, setHeight ] = React.useState(30)
+
+  console.log({ height })
```
lancetipton, line 34: This should be removed

repos/keg-components/src/hooks/useAutocompleteItems.js
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Returns a new array containing a subset of possibleValues, each of which is:
+ *  - unique; and
+ *  - either a substring of `text` or the same string.
+ * The filtering ignores casing and accents.
+ * @param {String} text - text to check (e.g. user input so far)
+ * @param {Array<string | Object>} possibleValues - string or object { text, key? } array
+ * @returns {Array<Object>} the new array of items, without duplicates
+ */
+export const getItemsMatchingText = (text, possibleValues) => {
```
lancetipton, line 50: I like it

repos/re-theme/src/reStyle/reStyle.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react'
-import { exists, noOpObj } from '@keg-hub/jsutils'
+import { exists } from '@keg-hub/jsutils'
 import { StyleInjector } from 'StyleInjector'
 import {
   getComponentName,
-  useObjWithIdentity,
+  useShallowMemoMerge,
```
lancetipton, line 6: I don't like it

repos/re-theme/src/reStyle/reStyleHooks.js
```diff
@@ -28,6 +28,23 @@ export const getComponentName = Component => {
...
 
+/**
+ * Hook that ensures a consistent Object identity to avoid re-renders
+ * @function
+ * @param {Object} identity - Object to be reused
+ * @param {Array} mergeObjs - Other objects to be merged with the identity object
+ *
+ * @returns {Object} - identity object merged with the passed in mergeObjs
+ */
+export const useObjWithIdentity = (identity, ...mergeObjs) => {
```
lancetipton, line 39: Is this still being used, or can we just remove it? Pretty sure it's only used in `reStyle`

lancetipton (changes requested):
Looks good. A couple things to look at. Let me know if you have questions. Thanks

repos/keg-components/src/components/form/scrollable/select/scrollableSelect.stories.mdx
mpcarolin: Hmm, not able to recreate this, but try the new image and let me know if you still encounter it.

repos/keg-components/src/components/form/scrollable/select/scrollableSelect.example.js
mpcarolin: updated

repos/re-theme/src/reStyle/reStyle.js
mpcarolin: lol

repos/re-theme/src/reStyle/reStyleHooks.js
mpcarolin: removed

repos/keg-components/src/components/form/autocomplete/autocomplete.example.js
mpcarolin: updated

repos/keg-components/src/components/form/scrollable/item/item.js
mpcarolin: Yep the `deepMerge` is done automatically within retheme. I updated the `item.example.js` story with custom styles to demonstrate this.

repos/keg-components/src/hooks/useAutocompleteItems.js
mpcarolin: 👍🏼

lancetipton (changes requested):
Looks good. Everything works as expected. I do have one recommendation.
When the list pops open
* My first instinct is to use the arrow keys to move up and down the list
* This is pretty common functionality and something we should include
* It also looks pretty straight forward to implement
  * [Here's an example](https://codesandbox.io/s/react-hooks-navigate-list-with-keyboard-eowzo)
 
I'm going to set this to request changes, but if you think this is too much for the ticket, let me know and I'll approve it and we can do it on another ticket.

Thanks

daniel112 (changes requested):
- I'm getting an error when trying to look at the `scrollableSelect`
    - [image](https://user-images.githubusercontent.com/3317835/108417208-ea2d7400-71ec-11eb-870c-97b5d35ecbc8.png)

- also do you mind adding the props to the storybook? for the Autocomplete and ScrollableSelect so we can see it in the storybook docs

- This is minor, but when there exists some empty spaces before i start typing, the autocomplete doesn't show up.  typically I would expect it to ignore the trailing spaces
    - [image](https://user-images.githubusercontent.com/3317835/108417764-a129ef80-71ed-11eb-9011-b8de33c63efc.png)

repos/keg-components/src/components/form/scrollable/select/scrollableSelect.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * A scrollable menu list of items, with prop-adjustable height and visibility
+ * @param {Object} props
+ * @param {Array} props.items - array of objects of the form { text, key } - key is optional if each item has unique text
+ * @param {Object} props.styles - style object for the view
+ * @param {Boolean} props.visible - boolean, indicating the menu is visible or hidden
+ * @param {Function} props.onSelect - the callback of form (item) => {...}. Fires when a menu item is selected
+ * @param {Number} props.height - the maximum height of the menu before scrolling is required to see remaining items.
+ */
+export const ScrollableSelect = ({
```
lancetipton, line 22: I'm really digging this component
We should update our Form/Select component to work with this. 
Way better then using the default browsers select dropdown

repos/keg-components/src/hooks/useKeyPress.js
```diff
@@ -0,0 +1,45 @@
...
+    [ setKeyPressed, targetKey ]
+  )
+
+  const upHandler = useCallback(
+    evt => evt.key === targetKey && setKeyPressed(false),
+    [ setKeyPressed, targetKey ]
+  )
+
+  useEffect(() => {
+    if (platform !== 'web') return
```
lancetipton, line 40: This is fine, I personally thing this flows better =>
```js
return platform === 'web'
  ? manageListeners(upHandler, downHandler)
  : noOp

```

repos/keg-components/src/hooks/useScrollIntoView.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {RefObject} ref - ref to element
+ * @param {boolean} shouldScroll - if true, keep ref.current visible
+ * @param {boolean|object} scrollOptions - options as defined at
+ *    https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameter
+ */
+export const useScrollIntoView = (ref, shouldScroll, scrollOptions=true) => {
+  useEffect(() => {
+    if (!ref.current)
+      return console.warn(
+        'Component does not appear to be accepting a ref. Unable to scroll it into view.',
```
lancetipton, line 17: This message seems odd, and  it possible we want to make setting the ref optional in some cases. I would just do this =>
```js
shouldScroll &&
  ref.current && 
  ref.current?.scrollIntoView(scrollOptions)
```

repos/re-theme/src/styleInjector/styleInjector.js
```diff
@@ -26,8 +26,8 @@ const BuildWithStyles = React.forwardRef((props, ref) => {
     <Component
       {...buildProps}
       style={filteredStyle}
-      ref={ref}
       className={isStr(Component) ? classList.join(' ') : classList}
+      ref={ref}
```
lancetipton, line 30: Odd changes to this file, but ok

lancetipton (changes requested):
Looks good, and works exactly as I would expect. 
Just a few small recommendations

daniel112 (approved):

repos/jsutils/src/number/mod.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {number} num 
+ * @param {number} divisor 
+ * @return {number?} the modulo result. Should be equivalent to
+ * return values from the `%` operator, except with negative `num` values.
+ * @example
+ * -1 % 10      // -1
+ * mod(-1, 10)  // 9
+ */
+export const mod = (num, divisor) => {
+  return ((num % divisor) + divisor) % divisor
```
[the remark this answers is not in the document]
mpcarolin: It's really useful whenever you want a wrap-around of a list. When the user presses the up arrow key when they are already at the top of the autocomplete list, for instance, this `mod` helps with wrapping around to the bottom of the list (the end of the array).

So: 
```autocompleteItems[mod(-1, autocompleteItems.length)]``` gives you the last element in the list
and 
```autocompleteItems[mod(autocompleteItems.length, autocompleteItems.length)]``` gives you the first element.

Python actually has this built in, you can just do ```some_array[-1]```; I always liked that.

File: repos/keg-components/src/components/form/autocomplete/autocomplete.example.js [hunk 13]
```diff
@@ -0,0 +1,45 @@
...
+  'Nigerian Dwarf',
+  'Oberhasli',
+  'Orobica',
+  'Saanen',
+  'Toggenburg',
+  'Russian White',
+  'Valais Blackneck',
+]
+
+export const Basic = () => {
```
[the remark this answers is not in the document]
mpcarolin: updated

repos/keg-components/src/hooks/useKeyPress.js
mpcarolin: updated

repos/keg-components/src/hooks/useScrollIntoView.js
mpcarolin: updated 👍🏼

repos/keg-components/src/hooks/useKeyPress.js
```diff
@@ -0,0 +1,47 @@
...
+    window.removeEventListener('keydown', downHandler)
+    window.removeEventListener('keyup', upHandler)
+  }
+}
+
+/**
+ * @param {string} targetKey
+ * @return {boolean} whether the key was pressed down or not
+ */
+export const useKeyPress = targetKey => {
```
lancetipton, line 27: I get why the hooks are before the final `useEffect` hook;
but I think it would be safe to move the `platform === 'web'` check up to the top of this method and here's why
* React requires the same amount of hooks to be called on every render relative to the components that are rendered
  * Which I'm pretty sure is the reason you put the `useCallback` hooks here   
* But, on native, we never want any of these hooks to run
* So if we more the `platform === 'web'` to the top of this hook, it should never call them on native.
  * Because they are never called, React doesn't care about them, and we should never have an issue if hooks being conditionally called

Personally, I find it easier to just separate the native and web hook versions into their own file
* This way you just let the resolver load the correct file, and you don't have to add extra logic into the hook to manage the platform
* It ends up being less code in the final bundle, and easier to reason about.
* It's just generally cleaner

I'm open to your thoughts on this, but I would prefer our other hooks such as `useClassList`, `useScroll`, and `useAnimate`

repos/keg-components/src/hooks/useOutsideDetect.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Listens for the user clicking outside the element stored in `ref`. If it detects a click,
+ * then it calls onOutsideClick.
+ * @param {RefObject} ref - ref object storing a DOM Node
+ * @param {Function?} onOutsideClick - fn of form (event) => {}
+ */
+export const useOutsideDetect = (ref, onOutsideClick) => {
+  useEffect(() => {
+    if (platform !== 'web')
```
lancetipton, line 24: See my comments above for the `useKeyPress` hook

repos/keg-components/src/hooks/useScrollIntoView.js
```diff
@@ -0,0 +1,25 @@
...
+ * is true.
+ * 
+ * @param {RefObject} ref - ref to element
+ * @param {boolean} shouldScroll - if true, keep ref.current visible
+ * @param {boolean|object} scrollOptions - options as defined at
+ *    https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameter
+ */
+export const useScrollIntoView = (ref, shouldScroll, scrollOptions=true) => {
+  useEffect(() => {
+    if (platform !== 'web') return
```
lancetipton, line 18: See my comments above for the `useKeyPress` hook

lancetipton (changes requested):
Everything works exactly how I'd expect. I left a few comments about refs.
Let me know if you have questions. Thanks

repos/keg-components/src/hooks/useKeyPress.js
mpcarolin: Totally agree. Updated to be consistent with the others.

repos/keg-components/src/hocs/withOutsideDetect.js
```diff
@@ -0,0 +1,34 @@
...
+      onOutsideClick, 
+      ...clientProps 
+    } = props
+
+    const ref = React.useRef()
+    useOutsideDetect(ref, onOutsideClick)
+
+    return <Component 
+      {...clientProps} 
+      ref={ref}
```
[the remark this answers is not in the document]
mpcarolin: Got it. Okay updated to accept a passed-in ref. Although I didn't need `handleRefUpdate`, unless I'm missing something

repos/keg-components/src/hooks/useOutsideDetect.js
mpcarolin: updated

repos/keg-components/src/hooks/useScrollIntoView.js
mpcarolin: updated

lancetipton (approved):
Looks good. I left a few comments but nothing blocking

repos/keg-components/src/hooks/useKeyPress/useKeyPress.js
```diff
@@ -0,0 +1,42 @@
+import { useState, useCallback, useEffect } from 'react'
```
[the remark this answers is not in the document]
mpcarolin: Agreed. I've been thinking that for awhile.

repos/keg-components/src/hooks/useOutsideDetect/useOutsideDetect.native.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Not implemented on native yet.
+ */
+export const useOutsideDetect = () => {
+  return console.warn('useOutsideDetect is not implemented on native platforms yet')
```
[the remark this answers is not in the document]
mpcarolin: updated

repos/keg-components/src/hooks/useScrollIntoView/useScrollIntoView.native.js
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Not yet implemented in native.
+ * @param {RefObject} ref - ref to element
+ * @param {boolean} shouldScroll - if true, keep ref.current visible
+ * @param {boolean|object} scrollOptions - options as defined at
+ *    https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameter
+ */
+export const useScrollIntoView = () => {
+  console.warn('useScrollIntoView is not implemented for native platforms yet.')
```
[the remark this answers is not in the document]
mpcarolin: updated

PR #140, opened 2021-02-26
Node Updates

mpcarolin:
## Context/Goal

* For my keg-herkin work, I wanted to add a node utility `tryRequire`
* Updates to allow the consumer to import the node utilities

## Updates

* `repos/jsutils/package.json`
  * updated the `files` property to include `src/node` 
* `repos/jsutils/src/node/tryRequire.js`
  * utility for safely requiring a module that might not exist, without throwing
* `repos/jsutils/src/node/cmd.js` and `repos/jsutils/src/node/loadModule.js`
  * just updated to ensure these use the `cjs` build files. Before, it was importing other jsutils es6 modules directly, and this was not compatible in my testing environment (keg-herkin tasks, which are vanilla node with no transpiling) 
* `repos/jsutils/src/node/index.js`
  * updated to export other modules from the cjs build folder instead of the untranspiled versions

## Testing

* `yarn test`
* The `tryRequire` utility has also already been tested in use in `keg-herkin`


repos/jsutils/src/node/tryRequire.js
```diff
@@ -0,0 +1,25 @@
...
+ * Tries to require the path, returning null if unable to.
+ * Does not throw.
+ * @function
+ * @param {string} filePath - path to file. You should use an absolute path
+ * @return {*?} the export at path, if it exists, null otherwise
+ * @example
+ * const module = tryRequire('/keg/tap/foo/bar.js')
+ * if (!module) console.log('bar.js module does not exist')
+ */
+const tryRequire = filePath => {
```
lancetipton, line 15: Can we make an async version of this as well. It's generally preferred when used cause it's not blocking.

It may not matter for what your doing with tests, but if used by a backend api it could lock things up.

Or if you your code can handle it as async that would best all around

repos/jsutils/package.json
```diff
@@ -15,7 +15,8 @@
   },
   "homepage": "https://github.com/simpleviewinc/keg-hub#readme",
   "files": [
-    "build"
+    "build",
+    "src/node"
```
lancetipton, line 19: It's so funny cause I was just thinking how we needed to do something like this.

I was planning to add a hook, to copy the node folder into the build folder, but this is so much better

repos/jsutils/src/node/cmd.js
```diff
@@ -2,13 +2,17 @@
 
 'use strict'
 
-const { logData, setLogs } = require('../log')
-const { promisify, wait } = require('../promise')
+const {
```
lancetipton, line 5: A number of the methods in this file are replaced by the span-Cmd repo.

Might be a good idea to remove any that are, to try and keep code size down

lancetipton: While some should be removed, it's not part of this ticket in any way, so only do it if you think it's worth it.

lancetipton (changes requested):
Looks good to me. Just a request for adding a non-blocking async version

Thanks

repos/jsutils/src/node/cmd.js
mpcarolin: Okay, I removed the `cmd` command since `spawn-cmd` supersedes it. But I didn't see any other functions replaced. I see `getArgs` has an identically named function in `spawn-cmd` but they are actually different.

repos/jsutils/src/node/tryRequire.js
mpcarolin: Okay I updated this:

`tryRequire` is now asynchronous.

`tryRequireSync` is a new function that is synchronous.

I tried to conform to the node naming scheme here.

repos/jsutils/src/node/cmd.js
lancetipton: Yah, getArgs was a very very early version of my idea for the args-parse repo.

So it is to getArgs what spawn-Cmd is to Cmd 

So really it could be removed as well

lancetipton: Actually, pretty sure this whole fine can be removed. 
All these functions are handled by other repos

Eventually what I would like to do is create a keg-utils repo, that moves a number of items from the the keg-cli as well as a few other place into one repo.

Then a number of the methods there could be moved into here, which would make them re-usable.

But that takes time like anything else. So for now just remove this file. It's not needed

mpcarolin: sounds good. It's removed

lancetipton (approved):
Looks good, thanks

PR #141, opened 2021-03-01
isQuoted helper

mpcarolin:
## Context

* I made some updates for `parkin`, and along the way I wrote this utility

## Updates

* `repos/jsutils/src/string/isQuoted.js`
  * checks if the string is wrapped in quotation marks

## Testing

* yarn test


lancetipton (approved):
looks good. All tests pass

PR #142, opened 2021-03-02
Regex and String Utils

mpcarolin:
## Context/Goal

* Needed yet more utils, this time specifically for regex work in `parkin`

## Updates

* `repos/jsutils/src/method/checkCall.js`
  * fixed bug/updated checkCall to always return the output of the function if that output is falsey.
  * previously, if the output was falsey, it would return `undefined` due to the `|| undefined` logic. For example, if the output of the method is`0`, then this was being hidden and replaced with `undefined`. Now, I made sure `checkCall` will return `0` for a function like that
  * this was the cause of a bug in `parkin`
* `repos/jsutils/src/regex/getRegexSource.js`
  * extracts the source regex text from input 
* `repos/jsutils/src/regex/isRegex.js`
  * checks if value is an instance of RegExp
* `repos/jsutils/src/regex/joinRegex.js`
  * combines regular expressions together within a single capture group
* `repos/jsutils/src/string/getWordEndingAt.js`
  * parses the word that **ends** at an index in a string
 * `repos/jsutils/src/string/getWordStartingAt.js`
   * parses the word that **starts** at an index in a string
* `repos/jsutils/src/string/reverseStr.js`
  * reverses a string

## Testing
* yarn test
* this will also be demonstrated as working in the PR I'm writing up now for parkin


repos/jsutils/src/regex/joinRegex.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {string} options - options string (the second argument of RegExp constructor)
+ * @example
+ * // calling using spread args
+ * const joined = joinRegex(/[A-z]+/, /[0-9]/, 'g')
+ * joined === /([A-z]+|[0-9])/g
+ * @example
+ * // calling with an array
+ * const joined = joinRegex([ ...allMyRegEx ], 'gi')
+ */
+export const joinRegex = (...args) => {
```
lancetipton, line 39: This one is odd, but I like it

lancetipton (approved):
All tests pass, and code looks good. 
There's some interesting things going on here.
I like the solutions you used, well done.

PR #152, opened 2021-04-22
Zen 609 automate package exec

lancetipton:
**Ticket**: [ZEN-609](https://jira.simpleviewtools.com/browse/ZEN-609)

* This PR should be reviewed along side [this keg-cli pr](https://github.com/simpleviewinc/keg-cli/pull/60)

> This is part 1 of the overall task.
> In this part, it automates the bundling of a taps code through a build action
> Which is then included in the image when running a `docker package` task
> Part two will handle running that bundle within the image

## Context

* When we run our docker packages on the staging server, there are a number of features that we don't need, like running in development modes
* These features slow down the execution, and in some cases prevent resolving the application url

## Goal
* Automate bundling the application when the docker image is created
* This will allow the staging server serve the bundle instead of using the development server

## Updates
* `repos/keg-components/package.json` && `repos/re-theme/package.json`
  * Added the `NODE_ENV` to all scripts that run storybook
  * Updated the `sb:server` script to match the command from the `serve action`
  * Update the `homepage` key, to allow expo bundles to be built properly
  * Added `build:optimize` script to optimize assets
* `repos/re-theme/container/values.yml` 
  * Added `build` and `serve` actions 
  * Added new DOC_BUILD_PATH env
* Updated ReThemes run.sh script to handle running a bundle in non-development environments

## Tests
* Do the tests in this PR => https://github.com/simpleviewinc/keg-cli/pull/60


repos/re-theme/container/values.yml
```diff
@@ -1,3 +1,13 @@
+actions:
+  tap:
+    build:
+      cmds:
+        - yarn sb:build
+        - rm -rf {{ envs.DOC_BUILD_PATH }}
+        - cp -R {{ envs.DOC_APP_PATH }}/docs {{ envs.DOC_BUILD_PATH }}
```
mpcarolin, line 7: Hmm I must be missing something: isn't this action building and then immediately deleting the build? Then it looks like it's just copying over the docs folder into the build path? Can you untangle this for me?

repos/re-theme/container/values.yml
```diff
@@ -1,3 +1,13 @@
...
+    serve:
+      cmds:
+        - npx serve {{ envs.DOC_BUILD_PATH }} --cors -n -l {{ envs.KEG_PROXY_PORT }}
```
mpcarolin, line 10: I really like how we can use those envs now

mpcarolin (commented):
Looks good. I left one comment referring to something confusing that I hope you can clear up for me.

repos/re-theme/container/values.yml
lancetipton: It could be that the ENVs are causing the confusion.
Look at it like this => 
* `DOC_APP_PATH/docs === /keg/tap/docs`
* `DOC_BUILD_PATH === /keg/tap-build`

Given that, if we follow the list of cmds =>
* `yarn sb:build` - creates a new build at `/keg/tap/docs`
  * Builds the application code in it's current state
* `rm -rf {{ envs.DOC_BUILD_PATH }}` - is the same as `rm -rf /keg/tap-build`
  * It removes the **OLD** build it one already exists
  * This is done to make sure we are starting fresh
  * This cmd could come first if we wanted it to
* `cp -R {{ envs.DOC_APP_PATH }}/docs {{ envs.DOC_BUILD_PATH }}` - is the same as `cp -R /keg/tap/docs /keg/tap-build`
  * It copies the **NEW** build created in the first command to the path we just deleted above

So basically it dose this
1. Create a new build of the current tap code
2. Remove the old build code by deleting the previously built folder ( if it exists )
3. Copy the new build to the same location as the old build we just deleted

mpcarolin (approved):
LGTM

PR #154, opened 2021-04-23
Migrate components to linked tap

lancetipton:
## Context

* Keg-Components should be a linked tap and work the same as retheme.
* Currently it's treated as an internal app, and make things more complicated especially for testing

## Goal

* Migrate keg-components to be a linked tap

## Updates

* Added container folder and files
* Updated dockerfile to work as a tap

## Testing

* TODO


mpcarolin (approved):
Looks good

PR #156, opened 2021-04-30
ZEN-543: Rebuild Workflow

mpcarolin:
**Ticket**: [ZEN-543](https://jira.simpleviewtools.com/browse/ZEN-543)

## Context/Goal

* To save developers from a lot of manual, error-prone work, we've decided to make a github workflow for building and pushing all of the keg-hub repos' docker images when a push to develop is made 

## Updates
* `.github/workflows/build-images.yml`
  * the workflow that clones and sets up the keg-cli and keg-hub, authentication, and builds and pushes each repo
* `scripts/ci/linkRepo.sh`
  * helper script that links a single injected tap repo
* `scripts/ci/linkTaps.js`
  * another helper that links all of them (currently: retheme & keg-components)

## Testing
* verify that this action succeeded: https://github.com/simpleviewinc/keg-hub/pull/156/checks?check_run_id=2472763766
  * this was run with the latest code for keg-cli and keg-hub, but feel free to run it again if you want, though it will take up to 30 minutes


.github/workflows/build-images.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  build-images:
+    name: 'Build Images'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    env:
+      CLI_BRANCH: ZEN-543-updates-for-workflow
```
lancetipton, line 17: Seems like this would be better as `${GITHUB_REF##*/}`
* Eventually we want this to run when a PR is submitted, to me it makes sense to make the branch dynamic
  * You probably planned to update it once approved, but it should work even now, so it's one less thing to worry about later

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,48 @@
...
+      run: git -C $GITHUB_WORKSPACE clone --single-branch --branch ${{env.CLI_BRANCH}} https://github.com/simpleviewinc/keg-cli.git
+    - name: "Install Keg-CLI"
+      id: install-cli
+      run: yarn --cwd $GITHUB_WORKSPACE/keg-cli install
+    - name: "Setup Keg-CLI"
+      id: setup-cli
+      run: cd $GITHUB_WORKSPACE/keg-cli; KEG_HUB_BRANCH=${GITHUB_REF##*/} KEG_ROOT_DIR=$GITHUB_WORKSPACE/keg-hub bash scripts/ci/setupCLIConfig.sh
+    - name: "Set Docker User"
+      id: set-docker-user
+      run: source $GITHUB_WORKSPACE/keg-cli/keg && keg config set key=docker.user value=${{env.CLI_USER}} --confirm false
```
lancetipton, line 33: Given that each step is inside the same job
  * Adding `source $GITHUB_WORKSPACE/keg-cli/keg` to each step should not be needed. 
  * I have not tested it, but each set should use the same environment, and therefore the file should already be loaded
  * You should just be able to call it once, and be good for all other steps

Another option, would be
  * Call the node script directly with => `node $GITHUB_WORKSPACE/keg-cli/keg-cli <params>`
  * For example, `line 33` would look like this =>
  `run: node $GITHUB_WORKSPACE/keg-cli/keg-cli config set key=docker.user value=${{env.CLI_USER}} --confirm false`

scripts/ci/linkTaps.js
```diff
@@ -0,0 +1,56 @@
+const { exec } = require('child_process');
```
lancetipton, line 1: This file along with the `scripts/ci/linkRepo.sh` file makes it overly complicated.

Why not just do everything in one file?
  * You could either call the `tap link` task directly from this file
  *  Or loop over the taps to link in the `scripts/ci/linkRepo.sh` file

In my opinion, it would be better to search the `keg-hub/repos/` sub folders
  * Within each sub-folder, check if it has a tap.js(on) file
  * If it does, then update the global config
    * By calling the `tap link` task
    * Or by calling `keg-cli/src/utils/globalConfig/addTapLink`
      * This is the method  the `tap link` task calls to update the global config

lancetipton (changes requested):
Looks good.
* I left a few comments that are based on my understanding of github actions/workflows
* While I could be wrong, I think it's worth validating to try and simplify the steps in the workflow

Thanks

.github/workflows/build-images.yml
lancetipton: Nevermind. This is for the keg-cli, not the keg-hub repo. Disregard this comment.

.github/workflows/build-images.yml
mpcarolin: Yeah I cleaned this up in the latest commit through refactoring.

But regarding the first option you presented, I actually tried that first but it did not work. [The docs elaborate](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idsteps): 
>Each step runs in its own process in the runner environment and has access to the workspace and filesystem. Because steps run in their own process, changes to environment variables are not preserved between steps. 

Regarding the second option, yeah that would be better (although I don't think it's all that different than just sourcing first), but should no longer be necessary.

scripts/ci/linkTaps.js
mpcarolin: Yeah all good points. 

I updated it to use one single script (`linkRepos.sh`).

That script loops over each repo in `keg-hub/repos`, and then it attempts to link that repo.

If that repo doesn't have a tap-config alias key, then it skips it. 

So we won't need to update this script in the future when we add more injected taps; it should just work.

scripts/ci/buildPushDockerImages.sh
```diff
@@ -0,0 +1,55 @@
...
+  keg_clean_docker
+
+  # only build one repo if the env is set
+  if [[ ! -z "$KEG_REPO" ]]; then
+    keg_build_repo $KEG_REPO $KEG_BUILD_TAG
+    return 
+  fi
+
+  # otherwise build and push each keg-hub image
+  for REPO in base core components tap retheme
```
lancetipton, line 48: With the updates to `keg-components` being moved to an external tap, it now depends on the `tap:develop` image.
* Currently in the action, `keg-components` is built before `tap`, which means it's using older version of it.
* `tap` should come before `components` here, to ensure the new version it built and pushed before `components` uses it to build it'd new version
* This ensure it's using the latest `tap:develop` image when building

scripts/ci/linkRepos.sh
```diff
@@ -0,0 +1,26 @@
...
+
+log_result () {
+  if [[ $2 -eq 1 ]]; then
+    echo "No alias key in tap config for \"$1\". Skipping..."
+  fi
+}
+
+link_repos() {
+  local KEG_REPOS=$(ls "$KEG_HUB_PATH/repos")
+  for REPO in $KEG_REPOS
```
lancetipton, line 14: Personally I like keeping the `do` part of the `for` loop on the same line
Not blocking, but for me it makes sense that they go together =>
```sh
for REPO in $KEG_REPOS; do
  # ...code stuff
done
```

lancetipton (changes requested):
Looks good. Left a couple of comments. Only 1 of them is blocking.

Let me know if you have questions. Thanks

scripts/ci/linkRepos.sh
mpcarolin: sure, updated

.github/workflows/keg-build-images.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  build-images:
+    name: 'Build Images'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    env:
+      GIT_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      KEG_CLI_BRANCH: ZEN-543-updates-for-workflow
```
lancetipton, line 18: I think you missed this one. I see line `33` was updated. Pretty sure this one should be too

scripts/ci/linkRepos.sh
lancetipton: For some reason, you're most recent commit is not showing up. but I do see the change in the commit directly. Odd

lancetipton (changes requested):
Just one small change in `.github/workflows/keg-build-images.yml` on line `18`

.github/workflows/keg-build-images.yml
mpcarolin: updated to support workflow_dispatch inputs as discussed

lancetipton (approved):
Looks good. Awesome work. Thanks man

PR #157, opened 2021-05-05
JSUtils mapFind

mpcarolin:
### Example
* Look at this example in `checkSubTasks`, in keg-cli at `src/utils/builders/buildGlobalTaskAlias`
```javascript
  return Object.keys(tasks).reduce((task, key) => {
    return task || get(tasks, `${ key }.tasks.${ command }`, task)
  }, null)
```
* here the function is reducing over an array of task names, getting the task command, and stopping only once it finds one that exists
* reduce still iterates over all keys
* then it's returning the value of the `get(...)` call

### mapFind
 * I think that example represents a common coding pattern we use: trying to find an element in array which, after a function maps that element, it passes a certain test function, and then we return that **mapped** value
    * we can't use `.find` for this because `find` returns the original element, not its mapped value
    * we could use `.map` and `.find` but that can be expensive
    * we often instead resort to using `.reduce` or a standard for loop
* we could simplify this with an abstraction like this:
```javascript
return mapFind(
    Object.keys(tasks),
    key => get(tasks, `${ key }.tasks.${ command }`)
  )
)
```
* I think this is simpler, less code, more readable
* and it can be made to iterate over elements *only until a match is found*, just like the `.find` array prototype function

## Updates
* `repos/jsutils/src/array/mapFind.js`
  * implemented the function described above
 * `*/.js`
   * added tests, exported the function

## Testing
* check that the tests pass


lancetipton (approved):
Looks good. All the tests pass

PR #179, opened 2021-08-09
Part 1 - fix(resolver): tap config load error

lancetipton:
**Ticket**: [ZEN-641](https://jira.simpleviewtools.com/browse/ZEN-614)

> **IMPORTANT** - This PR is just for reviewing code changes
> * All tests related to these changes should be preformed in [this PR](https://github.com/simpleviewinc/keg-hub/pull/190)

## Context
* The current alias namespace uses SV as the prefix
  * Every alias must use this prefix
  * It would be helpful to separate a keg-core alias from a taps alias
* At some point tap-resolver was updated to allow javascript files for the tap config
  * This works well, but javascript files load differently then json files
  * Previously if the javascript config threw an error while loading, it would get swallowed
  * This makes it very difficult to debug
  * Currently it will just load the package.json, and have all kinds of errors

## Goal
* Added code to throw an error when a taps config file can not be loaded
  * This fixes that issue by checking if the tap config file exists
  * If it exists, and it can not be loaded, then it throws an error letting the user know 
* Allow taps to have a separate alias prefix then keg-core
* Ensure taps are backwards compatible so they do not need to be updated

## Updates
* `repos/keg-core/app.json`
  * Add baseNamespace to keg-core
* `repos/tap-resolver/src/helpers/helpers.js`
  * Updated to show an error message when a tap config file exists, but can not be loaded 
* `repos/tap-resolver/src/builders/buildConstants.js`
  * Updated to add each alias under the `baseNamespace`
  * Because the alias is the same for both `nameSpace` and `baseNamespace`
    * Either alias can be used within the app and return the **same** result
    * This allows backwards compatibility, while also now allowing taps to define their own `namespace`
* `repos/tap-resolver/src/resolvers/contentResolver.js` 
  * Fixed a few outdated jsDoc comments

## Testing
* Go review [this PR](https://github.com/simpleviewinc/keg-hub/pull/190) and it's tests
* The tests in the PR use the changes in this PR


mpcarolin (approved):
Looks good

PR #184, opened 2021-08-12
chore(keg-core): Update dependencies 

lancetipton:
**Ticket**: [ZEN-614](https://jira.simpleviewtools.com/browse/ZEN-614)

## Goal

* Update the dependencies of keg-core
* Fix any issues related to the dependency updates

## Updates
* `repos/keg-core/core/base/assets/**`
  * Added a couple init files that are ignored by default
  * These files get overwritten by `tap-resolver`, but the build was throwing an error when they didn't exist
* `repos/keg-core/core/configs/eslintrc.config.js`
  * eslint config fixes due to dep updates
  * Fixed issue causing ESLint to error
    * Left comments
    * [See issue here](https://github.com/Intellicode/eslint-plugin-react-native/issues/270) for more info
  * Updated all the `dependencies` types to the latest versions where needed
  * Updated/Added `resolutions` to use the latest versions
  * Added missing `web:build` so it matches most tap build scripts

### Setup
* Run this command `keg dpg run --package keg-core:ZEN-614-update-core --env prod`

### Test 1
* Navigate to [the keg-core app](http://core-zen-614-update-core.local.keghub.io/) in your browser
  * Ensure the app runs as expected
* Use the select box in the top right to switch to the plugins page
  * Interact with the plugins page and ensure it works as expected
* You should not see any errors in the inspector console 

### Test 2
* Next validate that the keg-components and re-theme code is the correct code
* Connect to the running docker container `keg core att`
* Then run following command 
  * `cat /keg/keg-core/node_modules/@keg-hub/keg-components/package.json | grep @babel/core`
  * `cat /keg/keg-core/node_modules/@keg-hub/re-theme/package.json | grep @babel/core`
* Those commands should print out the `@babel/core` version from the package.json of the repo
  * Ensure the version is **7.14.8** for both
  * [Looks like this](https://i.imgur.com/TL5BhL8.png)
  * *Side note* - If you really want to, you can run `yarn build` in both `keg-components` and `re-theme` folders
    * This ensure the latest code is being used. But FYI, I already did that
* Compare that with the versions currently on develop
  * [keg-components](https://github.com/simpleviewinc/keg-hub/blob/8fa9013051f0bca13f0c0b162bf262f11b3c0a11/repos/keg-core/package.json#L51) 
  * [re-theme](https://github.com/simpleviewinc/keg-hub/blob/8fa9013051f0bca13f0c0b162bf262f11b3c0a11/repos/re-theme/package.json#L51)

### Test 3
* Run keg-core inside a tap by running `keg evf pack run evf:zen-614-tap-test` 
  * It should pull and run the image
*Navigate to [the events-force app] (http://evf-zen-614-tap-test.local.keghub.io/) in your browser
  * Ensure the tap works as expected
  * Ensure there are no issues related to running the tap, and loading the tap files
* Exit running container, then run the command => `keg evf pack run --package evf:zen-614-tap-test --env prod`
  * This should run the same container in production
  * Ensure the tap works as expected
  * Ensure there are no issues related to running the tap, and loading the tap files
* **Notes**
  * If you do find a bug, make sure it's related to the keg-core updates, and not the tap
  * You can use the [events-force demo](https://simpleviewinc.github.io/keg-test-consumer/) page to compare


mpcarolin (commented):
Looks good through Test 2. Test 3 is missing some details, though. Also one of the checks aren't passing.

mpcarolin (approved):
Looks good

PR #190, opened 2021-08-18
Part 2 - chore(core): Update alias namespace to use base

lancetipton:
**Ticket**: [ZEN-641](https://jira.simpleviewtools.com/browse/ZEN-614)

> **IMPORTANT** - This is part 2 of a two part PR.
> * If you have not already done so, go review [this PR](https://github.com/simpleviewinc/keg-hub/pull/179) first
> * All tests cover both this PR, and [part 1](https://github.com/simpleviewinc/keg-hub/pull/179)

> * **TESTS** - Currently the tests are failing, please see **Test 4** for more information

## Context

* The current alias namespace uses SV as the prefix. It also limits all alias to use the prefix
* It would be helpful to separate  a keg-core alias from a taps alias

## Goal
* Allow taps to have a separate alias prefix then keg-core
* Ensure taps are backwards compatible so they do not need to be updated

## Updates
* `repos/keg-core/app.json`
  * Add `baseNameSpace` property with the value of `Keg`
  * Add override, so the `baseNameSpace` can not be overwritten by a tap
  * Set the "namespace" to be `SV` so it's backwards compatible with previous taps
    * Unlike the `baseNamespace`, the `namespace` value can be overwritten by a tap
      * This in turn allows taps to set their own namespace prefix
* `repos/keg-core/**`
  * Updated all alias references to `SV` in keg-core to use the `Keg` prefix instead

### Test 1
* Run cmd `keg evf pack run evf:test-keg-core-alias`
* Navigate to the [events-force tap](http://evf-test-keg-core-alias.local.keghub.io/) in your browser
  * Ensure the tap works as expected
* Exit the container

### Test 2
* Switch to keg-hub => `keg hub`
* Pull this PR => `keg pr 190`
* Run the cmd => `keg evf start --sync core:install,resolver:install`
  * Navigate to the events-force app in your browser
    * The URL will depend on your current tap-events-force branch
  * Ensure the tap works as expected
* Leave the container running

### Test 3
* Open another terminal window
* Connect to the running tap with cmd => `keg evf att`
* Navigate to the keg-core folder => `cd node_modules/keg-core`
* Run the cmd => `cat app.json | grep ameSpace`
  * It should print out the namespace, and baseNamespace being used
  * [Looks like this](https://i.imgur.com/OPNqrGd.png)
* Ensure keg-core is using the update code from this PR

### Test 4
* The tests for keg-core are currently failing because of the changes to the `app.json`
  * This is because all the alias references in the files was updated to be `Keg`, but  the `namespace` value is set to `SV`
  * If we had the `tap-resolver` changes from [Part 1](https://github.com/simpleviewinc/keg-hub/pull/179) of this pr, then the tests would pass
  * So until the changes are merged upstream, and published to NPM, the tests in keg-core will fail
* To prove that the tests are still passing when the `tap-resolver` changes exist in node_modules
  * Connect to the running evf container => `keg evf att`
  * Copy keg-core to it's own folder => `cp -R tap/node_modules/keg-core /keg/keg-core`
  * Navigate into the new keg-core folder => `cd /keg/keg-core`
  * Run the tests =>  `yarn test`


mpcarolin (approved):
Looks good

PR #199, opened 2021-08-26
Universal Dynamic Styles

mpcarolin:
Ticket: https://jira.simpleviewtools.com/browse/ZEN-634

## Goal

* Bring `reStyle` and any component using `StyleInjector` to parity with theme files for dynamic styles
* Improve performance
* Simplify the theme-compiling code
* Add a few more shortcuts

### Overview
* Implemented `compileStyles`, which turned out to be a faster and simpler way to build styles both in the ReThemeProvider and in the StyleInjector
  * Updated it to be used everywhere
* Because it's in StyleInjector, `reStyle` and `keg-components` now have dynamic style capabilities

### Files
* `repos/re-theme/configs/eslintrc.config.js`
  * added some more exceptions for style aliases used in tests
* `repos/re-theme/src/constants/ruleHelpers.js`
  * Added a few more style shortcuts
* `repos/re-theme/src/context/reThemeProvider.js`
  * Updated to use `useCompiledStyles` for building the theme
  * I wanted to do this for consistency, but also because I determined this is actually 3-4x faster than the previous implementation
    * Old implementation avg. runtime: 28ms
    * Current implementation avg. runtime: 8.5ms
    * (Tested on Chrome, using tap-events-force, and switching between desktop and mobile themes about 20 times)
* `repos/re-theme/src/dimensions/sizeMap.js`
  * updated `getMergeSizes` to return the unused style keys in addition to the matching ones
  * Also updated the consumers of this function with the new return value
* `repos/re-theme/src/helpers/compileStyles.js`
  * The logic for compiling styles
  * Works recursively through the style object
  * This is inspired by `restructureTheme` and `buildTheme`, but it's simpler and actually runs faster
* `repos/re-theme/src/hooks/useCompiledStyles.js`
  * hook for calling `compileStyles`
  * gets the current state of the app (viewport size & platform), then calls compileStyles with those values
  * memoizes everything
  * adds meta properties too, if needed
* `repos/re-theme/src/helpers/getPlatforms.js`
  * helper to get currently active and inactive platforms
* `repos/re-theme/src/styleInjector/styleInjector.js`
  * updated to use `useCompiledStyles`
  * this enables universal dynamic styles: `reStyle` components, as well as injected components from `keg-components` can all use them
    * it works with props.style too

* `repos/re-theme/src/reStyle/reStyle.js`
  * updated `reStyle` to accept a second parameter for default props, e.g.:
    * I did this because sometimes styles are handled by props other than `style` and `styles`, and I think reStyle should be able to handle that
    * Also, sometimes the `reStyle`'d component is used only once, and it makes sense to specify the class name there
```js
reStyle(MySVGIcon)(styles, theme => ({ fill: theme.color.blue01, className: 'ef-foo' })
```
* `repos/re-theme/src/theme/getDefaultPlatforms.js`
   * just extracted this out of `buildTheme`, since it is used with `useCompiledStyles`
* `*.js`
  * I deleted files not used anymore
   
## Testing
* Verify all unit tests pass
* Run `keg evf pack run package=evf:zen-634-labels-move`
  * this starts up my tap-events-force branch, with this branch of re-theme synced in
  * the tap in the package is also updated to use style shortcuts and a platform key with reStyle
* Verify the tap looks and behaves normally

### ReStyle and StyleInjector
* Run any tap locally, with re-theme synced:
  * `keg evf start --sync retheme:install:start`
* Verify that you can pass in dynamic styles (e.g. `$small`, `$web`, shortcuts like `p: 10`) to:
  * any `reStyle` component
  * any injected component from `keg-components`, like `View`
* Verify the styles update correctly when you do


repos/re-theme/src/styleInjector/styleInjector.js
```diff
@@ -26,7 +27,7 @@ const BuildWithStyles = React.forwardRef((props, ref) => {
     <Component
       {...buildProps}
       style={filteredStyle}
-      className={isStr(Component) ? classList.join(' ') : classList}
+      className={classList.join(' ')}
```
lancetipton, line 30: Why was the `isStr` check removed? As far as I know this check is still needed?

repos/re-theme/src/dimensions/sizeMap.js
```diff
@@ -141,13 +141,20 @@ export const getSize = width => {
...
+ * const [ keys, unused ] = getMergeSizes('$medium')
+ * console.log(keys) -> [ '$medium', '$small', '$xsmall' ]
+ * console.log(unused) -> [ '$large' ]
  */
 export const getMergeSizes = key => {
-  // Add 1 because slice does not include the last item of the range
-  return sizeMap.entries
-    .slice(0, sizeMap.indexes[key] + 1)
-    .map(([ name, size ]) => name)
+  const keys = sizeMap.entries.map(([name]) => name)
```
lancetipton, line 155: Couldn't you just do `Object.keys(sizeMap.hash)` ?

lancetipton (changes requested):
Overall looks really good. 
There's a few items that need to be looked at, and or discussed. Let me know if you want to chat about them.
I havn't tested yet, as I'm anticipating a few more changes. Thanks

repos/re-theme/src/theme/getDefaultPlatforms.js
```diff
@@ -0,0 +1,25 @@
...
+
+// Default platforms to use when restructuring the theme
+// Use array, so we don't lose the order
+export const getDefaultPlatforms = () => {
+  const Platform = getRNPlatform()
+  // Rules for the OS platform ( web || ios || android )
+  const stylePlatforms = []
+
+  // If it's not a web platform, then add the $native platform
+  if (get(Platform, 'OS') !== 'web') stylePlatforms.push('$native')
```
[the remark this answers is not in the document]
mpcarolin: Updated

repos/re-theme/src/dimensions/sizeMap.js
mpcarolin: Ahh I meant to do this, but I forgot. Updated

lancetipton (approved):
Awesome work man. Thanks for doing this

PR #202, opened 2021-09-14
Update SvgIcon to support other elements than just Path

mpcarolin:
**Ticket**: [ZEN-632](https://jira.simpleviewtools.com/browse/ZEN-632)

## Goal

* ZEN-632 requires that I use the `Circle` svg element, but it is not supported by our `SvgIcon` wrapper component.
* We should export all the other elements of `react-native-svg` to allow consumers to use them, like:
```javascript
      <SvgIcon
        viewBox='0 0 20 20'
        size={20}
        className='keg-svg-children'
      >
        <SvgIcon.Circle 
          cx='10'
          cy='10'
          r='10'
          fill='#356C99'
        />
      </SvgIcon>
```

## Updates
* `repos/keg-components/src/components/svgIcon/svgIcon.examples.js` & `repos/keg-components/src/components/svgIcon/svgIcon.stories.mdx`
  * added an example
* `repos/keg-components/src/components/svgIcon/svgIcon.js`
  * adding the other elements to the `SvgIcon` export (e.g. `SvgIcon.Circle`)
* `repos/keg-components/src/components/svgIcon/svgIcon.native.js`
  * added children support
  * changed `<Path ...` to only render if `delta` prop is passed

## Testing
* `keg components pack run package=keg-components:svg-exports`
* Navigate to the `SvgIcon>With Children` story
* Verify that the circle icon renders

### DIFF
--- a/repos/keg-components/src/components/svgIcon/svgIcon.examples.js
+++ b/repos/keg-components/src/components/svgIcon/svgIcon.examples.js
@@ -75,3 +75,22 @@ export const Loading = props => {
     </StoryWrap>
   )
 }
+
+export const WithChildren = () => {
+  return (
+    <StoryWrap style={wrapStyles}>
+      <SvgIcon
+        viewBox='0 0 20 20'
+        size={20}
+        className='keg-svg-children'
+      >
+        <SvgIcon.Circle 
+          cx='10'
+          cy='10'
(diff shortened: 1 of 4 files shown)


lancetipton (approved):
Looks good. Thanks

PR #208, opened 2021-10-05
isIFrame Helper

mpcarolin:
**Ticket**: [ZEN-663](https://jira.simpleviewtools.com/browse/ZEN-663)

## Updates

* Add a helper to detect if running application is inside an iFrame

## Testing

* Verify tests have passed
* I already have this working in my tap-events-force branch, which will have a PR soon with tests inside an actual iframe

### DIFF
--- a/repos/keg-core/core/base/assets/index.js
+++ b/repos/keg-core/core/base/assets/index.js
@@ -4,5 +4,5 @@
 // ********************************* WARNING *********************************
 
 export const assets = {
-  ['clientLogo']: require('/Users/michael.carolin/keg-hub/repos/keg-core/core/base/assets/clientLogo.jpg')
+  ['clientLogo']: require('/home/runner/work/keg-hub/keg-hub/repos/keg-core/core/base/assets/clientLogo.jpg')
 }
--- a/repos/keg-core/core/base/assets/index.js
+++ b/repos/keg-core/core/base/assets/index.js
@@ -4,5 +4,5 @@
 // ********************************* WARNING *********************************
 
 export const assets = {
-  ['clientLogo']: require('/home/runner/work/keg-hub/keg-hub/repos/keg-core/core/base/assets/clientLogo.jpg')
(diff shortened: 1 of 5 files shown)


repos/keg-core/core/base/utils/platform/isIframe.js
```diff
@@ -0,0 +1,11 @@
+import { getWindow } from './getWindow'
+
+/**
+ * @returns {Boolean} true if this react app is in a web environment and also rendered inside of an iframe
+ */
+export const isIframe = () => {
+  const win = getWindow()
+  return win
+    ? win.location !== win.parent.location
```
lancetipton, line 9: While this check would work the majority of the time, it's technically not accurate.
  * Because both the location and the iframe could have the same location
  * A better check would be `window === window.parent`

repos/keg-core/core/base/assets/index.js
```diff
@@ -4,5 +4,5 @@
 // ********************************* WARNING *********************************
 
 export const assets = {
-  ['clientLogo']: require('/home/runner/work/keg-hub/keg-hub/repos/keg-core/core/base/assets/clientLogo.jpg')
+  ['clientLogo']: require('/Users/michael.carolin/keg-hub/repos/keg-core/core/base/assets/clientLogo.jpg')
```
lancetipton, line 7: This file should really be ignored in git, It seems odd this change was picked up.
That said, it seems you were running your changes on local, which is why it updated to your home path. 

Can you revert this file from the PR. Thanks

lancetipton (changes requested):
Take a look at my comments. Let me know if you have questions

repos/keg-core/core/base/utils/platform/isIframe.js
mpcarolin: Updated

repos/keg-core/core/base/assets/index.js
mpcarolin: Updated

lancetipton (approved):
Looks good, thanks man

